PR #34, opened 2023-04-01
Fix home manager module

FlafyDev:
From what I understand, the problem is that you can't override sets made with `neovimConfiguration`.

My solution was to get `neovimConfiguration` to `lib/module/default.nix` and just build a new package.
I did it by passing inputs to `lib/module/default.nix`, you might not want to do that, IDK.
I couldn't figure out how to add the default overlay (which contains `neovimConfiguration`) to the home manager module's pkgs (never worked with one). So i just passed inputs and got `neovimConfiguration` with that.

might also be a good idea to call it `mkNeovimConfiguration` instead?

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -45,7 +45,7 @@
 
         homeManagerModules.default = {
           imports = [
-            (import ./lib/module self.packages)
+            (import ./lib/module self.packages inputs)
           ];
         };
       };
--- a/lib/module/default.nix
+++ b/lib/module/default.nix
@@ -1,5 +1,5 @@
 # Home Manager module
-packages: {
+packages: inputs: {
   pkgs,
   config,
   lib ? pkgs.lib,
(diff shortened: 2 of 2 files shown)


NotAShelf:
> I did it by passing inputs to lib/module/default.nix, you might not want to do that, IDK.
frankly it's fine, making the thing *work* comes before following some nix dogma

> I couldn't figure out how to add the default overlay (which contains neovimConfiguration) to the home manager module's pkgs (never worked with one).
for the better, we do not want to pass the default overlay to the homeManager module specifically because it breaks useGlobalPkgs option of home-manager - you cannot pass nixpkgs options (i.e nixpkgs.overlays) when that option is set to true

> might also be a good idea to call it mkNeovimConfiguration instead?
neovimConfiguration sounds good to me, unless we make mkNeovimConfiguration a lib function that builds the package also outside the flake

PR #51, opened 2023-04-18
Dart fixes

FlafyDev:
fixes some things(see commits) with the Dart lsp to make it work.

- I tried to add an assert that prohibits dart.lsp.enable and dart.flutter-tools.enable to be both enabled, but i have no idea how to do it with nix's module system.

- I also changed the default from dart.lsp.enable to dart.flutter-tools.enable. The reason behind it is that flutter-tools is compatible with both dart and flutter lsps, while dart.lsp is not compatible with flutter. But I had noticed that flutter-tools requires dart(or flutter) to be in path while running nvim, while dart.lsp doesn't. So I can drop the commit if you want dart.lsp to be the default.

### DIFF
--- a/modules/languages/dart/config.nix
+++ b/modules/languages/dart/config.nix
@@ -12,7 +12,7 @@ with builtins; let
     dart = {
       package = pkgs.dart;
       lspConfig = ''
-        lspconfig.dart.setup{
+        lspconfig.dartls.setup{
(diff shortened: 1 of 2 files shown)


NotAShelf:
~~honestly I also have no idea how to use assertions~~

NotAShelf (approved):
LGTM!

PR #57, opened 2023-04-22
fix editorconfig option

horriblename:
the config string is written as lua instead of vimscript, and the logic is wrong anyways.

Changed the default to true because that's what neovim ships with, I could change it back if you want


NotAShelf:
Defaulting to true is fine, I set it to false by default as I don't use editorconfig and wouldn't be able to test/support it properly.

Thanks!

PR #59, opened 2023-04-23
fix: ccls config

horriblename:
ccls was not using the package defined in `cfg.lsp.package`

and also was using  `cfg.lsp.cclsOpts` which doesn't exist


NotAShelf:
Looks like another indev change slipped my mind, thank you.

While I will merge this, I think another module rework (especially to modules that call for a package and then get `/bin/<LSP>` from a package that not even provide that binary) is overdue.

NotAShelf (approved):

PR #63, opened 2023-04-26
feat: lazygit integration

horriblename:
https://github.com/NotAShelf/neovim-flake/issues/46

a simple lazygit thing using toggleterm, derived from their help file

### DIFF
--- a/modules/terminal/toggleterm/config.nix
+++ b/modules/terminal/toggleterm/config.nix
@@ -8,45 +8,52 @@ with builtins; let
   cfg = config.vim.terminal.toggleterm;
   toggleKey = "<c-t>";
 in {
-  config = mkIf cfg.enable {
-    vim.startPlugins = [
-      "toggleterm-nvim"
-    ];
+  config = mkMerge [
+    (mkIf cfg.enable {
+        vim.startPlugins = [
+          "toggleterm-nvim"
+        ];
 
-    vim.luaConfigRC.toggleterm = nvim.dag.entryAnywhere (''
-        require("toggleterm").setup({
-          open_mapping = [[${toggleKey}]],
-          direction = '${toString cfg.direction}',
-          -- TODO: this should probably be turned into a module that uses the lua function if and only if the user has not set it
-          size = function(term)
-            if term.direction == "horizontal" then
-              return 15
-            elseif term.direction == "vertical" then
-              return vim.o.columns * 0.4
+        vim.luaConfigRC.toggleterm = nvim.dag.entryAnywhere ''
+          require("toggleterm").setup({
+            open_mapping = [[${toggleKey}]],
+            direction = '${toString cfg.direction}',
+            -- TODO: this should probably be turned into a module that uses the lua function if and only if the user has not set it
+            size = function(term)
+              if term.direction == "horizontal" then
+                return 15
+              elseif term.direction == "vertical" then
+                return vim.o.columns * 0.4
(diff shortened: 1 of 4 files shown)


modules/terminal/toggleterm/config.nix
```diff
@@ -6,31 +6,47 @@
...
+      + optionalString cfg.lazygit.enable ''
+
+        local terminal = require 'toggleterm.terminal'
+        local lazygit = terminal.Terminal:new({
+          cmd = "lazygit",
+          direction = '${cfg.lazygit.direction}',
+          hidden = true,
+          on_open = function(term)
+            vim.cmd("startinsert!")
+            vim.keymap.set( 't', [[${toggleKey}]], function() term:toggle() end, {silent = true, noremap = true, buffer = term.bufnr})
```
NotAShelf, line 45: A separate dag here could be a better approach. Though, this is not exactly wrong, as it is still a part of toggleterm.

modules/terminal/toggleterm/toggleterm.nix
```diff
@@ -17,5 +17,13 @@ with builtins; {
       default = false;
       description = "Enable winbar";
     };
+    lazygit = {
+      enable = mkEnableOption "Enable LazyGit integration";
+      direction = mkOption {
+        type = types.enum ["horizontal" "vertical" "tab" "float"];
+        default = "float";
+        description = "Direction of the terminal";
```
NotAShelf, line 25: Clarify the terminal is for lazygit, "Direction of the lazygit window" would be more clear.

modules/terminal/toggleterm/config.nix
```diff
@@ -6,31 +6,47 @@
 with lib;
 with builtins; let
   cfg = config.vim.terminal.toggleterm;
+  toggleKey = "<c-t>";
 in {
   config = mkIf cfg.enable {
     vim.startPlugins = [
       "toggleterm-nvim"
     ];
```
NotAShelf, line 14: I believe lazygit requires, well, lazygit to be installed. Perhaps we can have a module that takes `nullOr package` as a type to optionally let the user bring their own lazygit.

NotAShelf (changes requested):

modules/terminal/toggleterm/config.nix
horriblename: better check this one, I barely understood what dag is

modules/terminal/toggleterm/config.nix
NotAShelf: You can use lib.optionals to append the package to the vim.startPlugins, which takes an input from the flake or a nixpkgs package. See `pluginsTypes` in `lib/types/plugins.nix`.

NotAShelf:
```nix
vim.startPlugins = [
    "toggleterm-nvim"
] ++ lib.optionals (lazygit.package != null) [
    cfg.lazygit.package
];
```

out of my head, this feels like a sane approach

NotAShelf:
LGTM! I'll update the docs and should be ready to merge afterwards.

NotAShelf:
Sorry it took me too long, had to go outside and touch grass for a bit. The alejandra formatting should fix the CI fail.

P.S. The styling CI looks like it's failing due to some Github networking error, Alejandra returns a positive exit code and I think we are good for a merge.

Thanks for your contribution.

PR #64, opened 2023-04-28
fix: broken python lsp option

horriblename:
python lsp option is written like as treesitter grammar


NotAShelf (approved):
LGTM

NotAShelf:
Sorry for the delayed response.

One would assume I wouldn't miss anything of such, but since there are not comments or notes I'm going to assume I did not do this on purpose.
Thanks for catching it!

PR #66, opened 2023-05-03
chore: upload docs build artifacts

n3oney:
This adds built docs available as .zip downloads for PRs and commits on the main branch, for easier checking when Nix(OS) isn't available.


NotAShelf (approved):
do we *really* want to upload all artifacts? it doesn't really hurt anyone to do so, but maybe it makes more sense to upload only the html

NotAShelf (approved):

NotAShelf:
Oh well, there is nothing to lose, so might as well.

PR #71, opened 2023-05-09
feat: new option to enable lua module loader

horriblename:
Enables the experimental lua loader in nvim 0.9, which caches modules like [impatient.nvim](https://github.com/lewis6991/impatient.nvim)

speeds up startup time quite a bit, for me it went from 300~500 ms down to 200 ms, tested with `nvim --startuptime`

### DIFF
--- a/modules/basic/module.nix
+++ b/modules/basic/module.nix
@@ -144,5 +144,7 @@ with builtins; {
       default = true;
       description = "Follow editorconfig rules in current directory";
     };
+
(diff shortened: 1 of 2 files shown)


horriblename:
I'm not sure how to put the code at the top of the `luaScript` section, but this probably isn't the best way?

NotAShelf (approved):

NotAShelf:
Thanks!

PR #73, opened 2023-05-10
feature: DAP support

horriblename:
Add support for debug adaptor protocol (DAP) via nvim-dap.

## Note

- ~~nvim-dap-ui is kinda broken in nvim-0.9, but will be fixed in 9.1: https://github.com/rcarriga/nvim-dap-ui/issues/262~~ fixed in nvim 0.9.1

- [x] nvim-dap
  - [x] custom keybinds
- [x] nvim-dap-ui
  - [x] custom keybinds
- [ ] languages
  - [ ] dart
  - [ ] elixir
  - [x] clang
  - [x] go
  - [x] python
  - [x] rust
  - [ ] ts
  - ~~[ ] html~~
  - ~~[ ] nix~~
  - ~~[ ] sql~~
  - ~~[ ] markdown~~
  - ~~[ ] tidal~~
  - ~~[ ] zig~~

I'll add basic configurations for supported languages, but I can only test go, python, rust and c/++.

## References

[nvim-dap's Debug Adapter Installation wiki](https://github.com/mfussenegger/nvim-dap/wiki/Debug-Adapter-installation)

### DIFF
--- a/modules/languages/python.nix
+++ b/modules/languages/python.nix
@@ -36,6 +36,69 @@ with builtins; let
       '';
     };
   };
+
+  defaultDebugger = "debugpy";
+  debuggers = {
+    debugpy = {
+      # idk if this is the best way to install/run debugpy
+      package = pkgs.python3.withPackages (ps: with ps; [debugpy]);
+      dapConfig = ''
+        dap.adapters.python = function(cb, config)
+          if config.request == 'attach' then
+            ---@diagnostic disable-next-line: undefined-field
+            local port = (config.connect or config).port
+            ---@diagnostic disable-next-line: undefined-field
+            local host = (config.connect or config).host or '127.0.0.1'
+            cb({
+              type = 'server',
+              port = assert(port, '`connect.port` is required for a python `attach` configuration'),
+              host = host,
+              options = {
+                source_filetype = 'python',
+              },
+            })
+          else
+            cb({
+              type = 'executable',
+              command = '${getExe cfg.dap.package}',
+              args = { '-m', 'debugpy.adapter' },
+              options = {
+                source_filetype = 'python',
+              },
+            })
+          end
+        end
+
+        dap.configurations.python = {
+          {
+            -- The first three options are required by nvim-dap
+            type = 'python'; -- the type here established the link to the adapter definition: `dap.adapters.python`
+            request = 'launch';
+            name = "Launch file";
+
+            -- Options below are for debugpy, see https://github.com/microsoft/debugpy/wiki/Debug-configuration-settings for supported options
+
+            program = "''${file}"; -- This configuration will launch the current file if used.
+            pythonPath = function()
+              -- debugpy supports launching an application with a different interpreter then the one used to launch debugpy itself.
+              -- The code below looks for a `venv` or `.venv` folder in the current directly and uses the python within.
+              -- You could adapt this - to for example use the `VIRTUAL_ENV` environment variable.
+              local cwd = vim.fn.getcwd()
+              if vim.fn.executable(cwd .. '/venv/bin/python') == 1 then
+                return cwd .. '/venv/bin/python'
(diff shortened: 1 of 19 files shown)


NotAShelf:
I'll try to help with testing as much as I can. Can test JS/TS and probably can try dart. Elixir support is a hit or miss anyway, so we can cross that bridge when we get to it.

horriblename:
I'm giving up on dart, elixir and ts:

- flutter-tools doesn't seem to work at all
- I have no clue what I'm doing in elixir
- I need to package [vscode-js-debug](https://github.com/microsoft/vscode-js-debug) for js

If anyone wants dart and ts support open a new issue and I'll try again in the future. not elixir though, I have no clue how to do that.

Other stuff:
- clang and rust uses lldb currently, I'm planning to replace it with the codelldb wrapper in the future to get better symbols in C++/rust

For future reference:
- codelldb attempt: https://github.com/horriblename/neovim-flake/commit/c2fb48e4eeac87b8c83345b03dbd7fcf10d44568
- elixir attempt: https://github.com/horriblename/neovim-flake/commit/b5c66b5bca45e5251e262240a878b805496a51d5
- dart attempt: https://github.com/horriblename/neovim-flake/commit/c3531e9a1960e51d7ec76c0bee6a9849829e8d3b

horriblename:
> should I merge this right away after I review it?

merge it right away. 

It's currently _usable_ but some symbols in C++/rust may show up as giberrish (e.g. `String` showing up as a memory address instead of the string itself)

FlafyDev:
I remember flutter-tools.nvim worked with dap when I used it.
What have you tried to do to make it work? 
And did you read https://github.com/akinsho/flutter-tools.nvim#debugging ?

horriblename:
flutter-tools doesn't work at all because it can't find dart, long story short flutter-tools tries to find dart in the same directory as the `flutter` binary but dart doesn't exist in wrapped flutter. [flafy's patch](https://github.com/FlafyDev/nixos-config/blob/main/modules/programs/neovim/flutter-tools-no-resolve.patch) might fix it (thanks flafy)

will try the patch over the weekend and see if it works

NotAShelf:
Spot on, I was waiting until monday to come so that I would be able to review properly. Keep me posted please!

modules/languages/python.nix
```diff
@@ -36,6 +36,69 @@ with builtins; let
       '';
     };
   };
+
+  defaultDebugger = "debugpy";
+  debuggers = {
+    debugpy = {
+      # idk if this is the best way to install/run debugpy
+      package = pkgs.python3.withPackages (ps: with ps; [debugpy]);
```
NotAShelf, line 44: is there a reason we provide python3 alongside the debugpy package? [python311Packages.debugpy](https://search.nixos.org/packages?channel=23.05&show=python311Packages.debugpy&from=0&size=50&sort=relevance&type=packages&query=debugpy) is already packaged for nixpkgs

NotAShelf (approved):

NotAShelf:
Overall looks good, I appreciate your time and effort!

I believe (without extensive testing) that this is ready for a merge, if you could update the release notes, that would be great.

modules/languages/python.nix
horriblename: I simply don't know of a better way. the debugpy package provides the module but its not executable, you need a python binary to use it: `python -m debugpy`. I could in theory add debugpy to PYTHON_PATH and the user can use whatever python binary they have, but, too complicated.

NotAShelf: Fair point, let's keep it as is for the time being.

horriblename:
added release notes, and dart support. you should merge the flutter-tools fix first, I rebased against that.

can you test the dart dap? just go into a dart file, set a breakpoint with `<space>db` then start debugging with `<space>dc`. For some reason my hello world app can't compile anymore even though it worked last week :c

NotAShelf:
Thanks!

PR #77, opened 2023-05-12
fix formatting

### LINKED ISSUE #62: Auto format messing up file
### ⚠️ Please verify that this bug has NOT been reported before.

- [X] I checked all existing issues and didn't find a similar issue

### Description

Sometimes the formatting messes up and deletes part of the file, usually when there is an error (syntax error etc. in the code detected by lsp), this happens with clangd sometimes, and throws this error:

```
Error executing vim.schedule lua callback: ...-unwrapped-0.9.0/share/nvim/runtime/lua/vim/lsp/util.lua:501: Invalid 'start_col': out of range
stack traceback:
	[C]: in function 'nvim_buf_set_text'
	...-unwrapped-0.9.0/share/nvim/runtime/lua/vim/lsp/util.lua:501: in function 'apply_text_edits'
	...rapped-0.9.0/share/nvim/runtime/lua/vim/lsp/handlers.lua:318: in function 'handler'
	...eovim-unwrapped-0.9.0/share/nvim/runtime/lua/vim/lsp.lua:1394: in function ''
	vim/_editor.lua: in function <vim/_editor.lua:0>
```

I looked at the source and this is how you're doing the formatting:
```
            if vim.g.formatsave then
              if client.supports_method("textDocument/formatting") then
                local params = require'vim.lsp.util'.make_formatting_params({})
                client.request('textDocument/formatting', params, nil, bufnr)

              end
            end
```

I have frankly never seen this, I'm guessing it's really old code? Neovim provides `vim.lsp.buf.format()` which doesn't cause the same problem, so I think it's better to switch to that

### 👟 Reproduction steps

not sure if this is easily reproduced:
1. use clangd
2. write a c++ file with some kind of obvious syntax error (e.g. missing `;`)
3. save

### 👀 Expected behavior

format normally. 

Running `:lua vim.lsp.buf.format()` to format it manually first fixes the issue

### 😓 Actual Behavior

neovim throws an error and the formatting is messed up, possibly deleting parts of the code.

### 💻 Metadata

 - system: `"x86_64-linux"`  - host os: `Linux 6.2.10-zen1-1-zen, Arch Linux, noversion, rolling`  - multi-user?: `yes`  - sandbox: `yes`  - version: `nix-env (Nix) 2.14.1`  - channels(py): `"home-manager, nixpkgs"`  - nixpkgs: `/home/py/.nix-defexpr/channels/nixpkgs`

### 📝 Relevant log output

```bash
Error executing vim.schedule lua callback: ...-unwrapped-0.9.0/share/nvim/runtime/lua/vim/lsp/util.lua:501: Invalid 'start_col': out of range
stack traceback:
	[C]: in function 'nvim_buf_set_text'
	...-unwrapped-0.9.0/share/nvim/runtime/lua/vim/lsp/util.lua:501: in function 'apply_text_edits'
	...rapped-0.9.0/share/nvim/runtime/lua/vim/lsp/handlers.lua:318: in function 'handler'
	...eovim-unwrapped-0.9.0/share/nvim/runtime/lua/vim/lsp.lua:1394: in function ''
	vim/_editor.lua: in function <vim/_editor.lua:0>
```
```

n3oney:
Should fix #62, but we need to have people test it.

### DIFF
--- a/modules/lsp/config.nix
+++ b/modules/lsp/config.nix
@@ -36,20 +36,21 @@ in {
       end
 
       -- Enable formatting
-      format_callback = function(client, bufnr)
-        vim.api.nvim_create_autocmd("BufWritePre", {
-          group = augroup,
-          buffer = bufnr,
-          callback = function()
-            if vim.g.formatsave then
-              if client.supports_method("textDocument/formatting") then
-                local params = require'vim.lsp.util'.make_formatting_params({})
-                client.request('textDocument/formatting', params, nil, bufnr)
+      local augroup = vim.api.nvim_create_augroup("LspFormatting", {})
 
-              end
-            end
-          end
-        })
+      format_callback = function(client, bufnr)
+        if vim.g.formatsave and client.supports_method("textDocument/formatting") then
+          vim.api.nvim_clear_autocmds({ group = augroup, buffer = bufnr })
+          vim.api.nvim_create_autocmd("BufWritePre", {
+            group = augroup,
+            buffer = bufnr,
+            callback = function()
+              vim.lsp.buf.format({
+                bufnr = bufnr
+              })
+            end,
+          })
+        end
       end
 
       default_on_attach = function(client, bufnr)


horriblename:
Doesn't do anything in c++ (clangd), if I read correctly, you filtered out non null-lsp sources?

n3oney:
> Doesn't do anything in c++ (clangd), if I read correctly, you filtered out non null-lsp sources?
> 
> 
> 
> I think you want to filter out tsserver? you should do exactly that `client.name ~= 'tsserver'`

Hm. I assumed we should do *all* formatting through null-ls. Wouldn't it be better for us to add clang_format as a null-ls source?
Makes it more centralized.

horriblename:
it feels kinda weird to make null-ls a requirement for formatting when LSPs can do it themselves tho.

just a heads up, `clang-format` is "part of" clangd (`nixpkgs.clang-tools`) so you'll either have to use something else, or make `clang-tools` mandatory for c/++

n3oney:
Fair point. I'll try to make it work so it *prefers* null-ls, while still allowing others when null-ls isn't available.

n3oney:
Actually, since it worked before the filter, no reason to add it now.

n3oney:
pls try update @horriblename

horriblename:
works without issue, thanks :+1:

NotAShelf:
Thank you both for your contributions!

NotAShelf (approved):
LGTM

PR #82, opened 2023-05-23
feat: give null-ls priority over other formatters

n3oney:
this PR fixes an issue I've had with formatting - when coding in TS, prettier would format the code fine, but so would tsserver, resulting in prettier being useless.
Before, I've tried to just filter every formatter that isn't null-ls to fix that issue, but that is not a good workaround as some languages don't have null-ls for formatting.

This pr introduces a better solution - it filters all formatters except null-ls, only when null-ls *can* format the file, which is pretty much the behaviour you'd expect.

*@horriblename could you test if this works fine for you? Thank you.*

### DIFF
--- a/docs/release-notes/rl-0.4.adoc
+++ b/docs/release-notes/rl-0.4.adoc
@@ -20,7 +20,11 @@ https://github.com/n3oney[n3oney]:
 
 * Refactored the resolveDag function - you can just provide a string now, which will default to dag.entryAnywhere
 
-* Fixed formatting issues and gave null-ls priority over other formatters
+* Fixed formatting sometimes removing parts of files
+
+* Made formatting synchronous
+
+* Gave null-ls priority over other formatters
 
 https://github.com/horriblename[horriblename]:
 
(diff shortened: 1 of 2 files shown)


docs/release-notes/rl-0.4.adoc
```diff
@@ -20,6 +20,8 @@ https://github.com/n3oney[n3oney]:
 
 * Refactored the resolveDag function - you can just provide a string now, which will default to dag.entryAnywhere
 
+* Fixed formatting issues and gave null-ls priority over other formatters
+
```
NotAShelf, line 24: I'd say this is ready for a merge after proper testing, but please do elaborate on "formatting issues" here.

NotAShelf (approved):

docs/release-notes/rl-0.4.adoc
n3oney: i put it there for #77

horriblename:
no problems on my end

docs/release-notes/rl-0.4.adoc
NotAShelf: As in within the documentation

n3oney: what

NotAShelf (approved):

Issue #85, opened 2023-06-11
null-ls cannot be used with ccls due to `offset_encoding`


NotAShelf:
I've recently come to realize that we cannot use `null-ls` with ccls (which is the default lang server for C/C++ rn) due `offset_encoding` differences. 

Thanks to this PR (#61) we are able to use clangd as the the language server, which will throw constant warnings for most of my C/C++ files.

```warning: multiple different client offset_encodings detected for buffer, this is not supported yet```

I've quickly attempted the suggested "fix" from [here](https://github.com/jose-elias-alvarez/null-ls.nvim/issues/428#issuecomment-1120578988) for null-ls, but doesn't seem to affect anything.

We should consider the following:
1. disable null-ls for ccls (@n3oney this might be more for you)
2. configure offset encodings for clangd to supress the multiple different client offset_encodings detected warning (@horriblename maybe you might know anything about this)

Relevant:

* Null-ls issue: https://github.com/jose-elias-alvarez/null-ls.nvim/issues/428
* Context around ccls: https://github.com/neovim/nvim-lspconfig/pull/1649
* Neovim issue: https://github.com/neovim/neovim/pull/16694

horriblename:
I don't understand, clangd works fine now right? why should it be changed?

NotAShelf:
I'm not sure what's wrong for me then, I see this for quite literally ever .cpp file I open:
![image](https://github.com/NotAShelf/neovim-flake/assets/62766066/857cf4f2-62d2-4da3-9906-7fb3999ad20c)

horriblename:
ah I'm guessing it's cos I disabled the git code action thing. alr, I'll look into it, should be an easy fix

horriblename:
For ccls, I think filtering out the warnings is probably better than disabling null-ls. afaik `vim.git.gitsigns.codeActions` is the one using null-ls here.

From the null-ls issue:

> This is a legitimate warning, but if you don't care and just want to suppress it, you can filter it out by overriding vim.notify:
> ```lua
> local notify = vim.notify
> vim.notify = function(msg, ...)
>     if msg:match("warning: multiple different client offset_encodings") then
>         return
>     end
> 
>     notify(msg, ...)
> end
> ```

NotAShelf:
I'm keeping this open for the time being, as any attempt at filtering the message (be it through vim.notify or nvim-notify) seems to fail.

PR #86, opened 2023-06-11
fix(clangd): use same offsetEncoding as null-ls

horriblename:
#85

fixes "multiple different client offset_encodings" warning when using clangd with null-ls


NotAShelf (approved):
LGTM

NotAShelf:
I think I'll also switch the default LSP to clangd after this, looks like it's working better than ccls.

NotAShelf:
Thanks!

PR #87, opened 2023-06-25
fix(flutter-tools): bug where dart lsp is not found

horriblename:
Fixes a bug where flutter-tools can't find dart binary when using a wrapped flutter package.

Patch taken from <https://github.com/FlafyDev/nixos-config>

### DIFF
--- a/docs/release-notes/rl-0.4.adoc
+++ b/docs/release-notes/rl-0.4.adoc
@@ -1,6 +1,5 @@
 [[sec-release-0.4]]
 == Release 0.4
-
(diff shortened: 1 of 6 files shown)


horriblename:
this fixes use cases where flutter was installed via nix, but would probably break cases where flutter was installed from some other source, perhaps I should make this an option?

NotAShelf:
> this fixes use cases where flutter was installed via nix, but would probably break cases where flutter was installed from some other source, perhaps I should make this an option?

Sorry, I just saw this comment.
I think we should assume, for the time being, that this flake is used explicitly on NixOS. I intend to release 0.4 soon, but after 0.4 release we can look at QoL changes such as making it optional. I'll be in touch when the time comes to take a look.

horriblename:
actually I already added the option `dart.flutter-tools.enableNoResolvePatch` which defaults to true

PR #95, opened 2023-07-19
feature/breadcrumbs

NotAShelf:
Introduces a breadcrumbs section on the Winbar (through lualine) and adds Navbuddy, a convenient UI for interacting with breadcrumb elements.

Mostly configurable, missing keybind and icons configuration as well as some strings that take tables.

TODO:
- [x] make keybinds configurable
- [x] make icons configurable
~~alternative breadcrumbs impl through [dropbar.nvim](https://github.com/Bekaboo/dropbar.nvim)~~[^1]

[^1]: requires neovim unstable, which we don't base off of


NotAShelf:
Practically ready to be merged, but will clean up the module and add nix options for each lua configuration option before I merge.

Breadcrumbs example (winbar)
image

Breadcrumbs example (navbuddy UI)
image

PR #99, opened 2023-07-21
Feature/extra plugins

horriblename:
https://github.com/NotAShelf/neovim-flake/issues/94

### DIFF
--- a/lib/types/plugins.nix
+++ b/lib/types/plugins.nix
@@ -89,20 +89,20 @@ with lib; let
     "nvim-dap-ui"
   ];
   # You can either use the name of the plugin or a package.
-  pluginsType = with types;
-    listOf (
-      nullOr (
-        either
-        (enum availablePlugins)
-        package
-      )
+  pluginType = with types;
+    nullOr (
+      either
+      package
+      (enum availablePlugins)
     );
+
+  pluginsType = types.listOf pluginType;
 
   extraPluginType = with types;
     submodule {
       options = {
         package = mkOption {
-          type = pluginsType;
+          type = pluginType;
         };
         dependencies = mkOption {
           type = listOf str;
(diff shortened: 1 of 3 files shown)


lib/types/plugins.nix
```diff
@@ -89,15 +89,34 @@ with lib; let
...
-      nullOr (
-        either
-        (enum availablePlugins)
-        package
-      )
+  pluginType = with types;
+    nullOr (
+      either
+      package
+      (enum availablePlugins)
```
NotAShelf, line 96: This has the potential to be a breaking change, did you test it thoroughly?

modules/core/default.nix
```diff
@@ -158,6 +158,11 @@ in {
       description = "List of plugins to optionally load";
     };
 
+    extraPlugins = mkOption {
+      type = types.attrsOf nvim.types.extraPluginType;
+      default = {};
+    };
```
NotAShelf, line 164: Wouldn't hurt to add a description here.

lib/types/plugins.nix
horriblename: I just extracted `pluginsType` into `pluginsType` + `pluginType`, cuz `extraPlugins.<name>.package` wants a single package instead of a list, I didn't change anything

horriblename:
renamed `dependencies` into `after` cuz dependencies sounds like we're doing package management stuff when we're not, we're just writing lua code in the right order

lib/types/plugins.nix
NotAShelf: just wanted to make sure it's not referenced in somewhere obscure that we might've missed, but it seems like you've already accounted for that possibility

marking as resolved

NotAShelf:
The implementation looks good. I'll daily drive this for a day or two to make sure everything works as intended before merging.

NotAShelf:
Okay, it looks good. I'll update the documentation shortly, and then merge. Thank you!

PR #100, opened 2023-07-21
fix: jittery scrolling due to flickering winbar

FrothyMarrow:
Prevent the buffer content from jumping around when scrolling because of disappearing lualine winbar 

https://github.com/NotAShelf/neovim-flake/assets/76622149/b1b91495-d2a4-4ce5-9de2-91ac5b52ae51


NotAShelf:
Looks good, but I'd rather make it optional for the sake of configurability. 

Will push a commit shortly.

NotAShelf (approved):

PR #128, opened 2023-08-14
fix: use boolToString on boolean

horriblename:
boolean values should use boolToString instead


NotAShelf (approved):
Good catch, thanks!.

PR #129, opened 2023-08-17
feat: add mkMultipleMappingOption

jacekpoz:
same as mkMappingOption but allows for multiple keybinds


jacekpoz:
turns out I can't `inherit defaults` because there's no `defaults` in `mkOption`

should I just change the name to `default` or `default = defaults;`?

NotAShelf (approved):

PR #142, opened 2023-09-27
languages/bash: add LSP

NotAShelf:
Adds a language server for bash.

### DIFF
--- a/configuration.nix
+++ b/configuration.nix
@@ -72,7 +72,7 @@ inputs: let
         python.enable = isMaximal;
         dart.enable = isMaximal;
         elixir.enable = false;
-        bash.enable = true;
+        bash.enable = isMaximal;
       };
 
       vim.visuals = {
--- a/configuration.nix
+++ b/configuration.nix
@@ -72,6 +72,7 @@ inputs: let
         python.enable = isMaximal;
         dart.enable = isMaximal;
         elixir.enable = false;
+        bash.enable = true;
       };
 
       vim.visuals = {
--- a/modules/languages/bash/config.nix
+++ b/modules/languages/bash/config.nix
@@ -1,0 +1,80 @@
+{
+  pkgs,
+  config,
+  lib,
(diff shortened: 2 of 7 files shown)


configuration.nix
```diff
@@ -72,6 +72,7 @@ inputs: let
         python.enable = isMaximal;
         dart.enable = isMaximal;
         elixir.enable = false;
+        bash.enable = true;
```
horriblename, line 75: should this be `isMaximal`? in my opinion bash is not "important" enough to install in minimal config

```suggestion
        bash.enable = isMaximal;
```

NotAShelf: That is a fair point. I guess it's a nice to have since it's basically the lowest barrier of entry language we have support for, but not really necessary.

modules/languages/bash/config.nix
```diff
@@ -0,0 +1,80 @@
...
+          on_attach = default_on_attach;
+          cmd = ${
+          if isList cfg.lsp.package
+          then nvim.lua.expToLua cfg.lsp.package
+          else ''{"${cfg.lsp.package}/bin/bash-language-server",  "start"}''
+        };
+        }
+      '';
+    };
+  };
```
horriblename, line 53: I don't like that these vars are duplicated in both files - it's easy to forget one when editing the other later on. maybe extract into another file and import it?

anyways, flutter module has the same problem and should be fixed as well.

horriblename (changes requested):

modules/languages/bash/config.nix
NotAShelf: I'll probably leave fixing this to a later revision, preferably one that focuses on cleaning up the language module as a whole. Ideally we'd do something that goes roughly as

```
languages/languageName
├── commons.nix # where formatters and servers would be
├── config.nix # config
├── default.nix # imports
└── languageName.nix # option declarations
```

PR #145, opened 2023-10-04
Revert "fix: stop using deprecated buildVimPluginFrom2Nix"

horriblename:
Reverts NotAShelf/neovim-flake#144 due to plugins not building


horriblename:
oops the problem is fixed by bumping nixpkgs, closing

PR #152, opened 2023-10-11
CI: add more stict checking

NotAShelf:
Introduces an editorconfig configuration to enforce standards across the source tree as well as a github workflow to check the validity of new files on pull requests.


NotAShelf:
alright does formatting not work

PR #154, opened 2023-10-14
modules/visuals: update indent-blankline to v3

NotAShelf:
Updates indent-blankline to v3 and removes the workaround requiring `vim.wo.colorcolumn = "99999"` - reported to be fixed upstream in neovim.

I've renamed some of the module options, and I'd *really* like to use `lib.mkRemovedOptionModule` and `lib.mkRenamedOptionModule`, however, both of those cause the build system to assert the files they are defined in (wherever they are) are not actually modules.

### DIFF
--- a/configuration.nix
+++ b/configuration.nix
@@ -90,7 +90,7 @@ inputs: let
           fillChar = null;
           eolChar = null;
           scope = {
-            showCurrContext = true;
+            enabled = true;
           };
         };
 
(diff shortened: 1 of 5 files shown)


NotAShelf:
Blocked by #178

FrothyMarrow:
I think `scope.enabled` might be a better name for `scope.showCurrContext` and helps users that are translating examples from README.md. For same reasons, description might be better defined as `Current scope of variables and functions with treesitter`. Other than that it looks good to me 👍

FrothyMarrow (approved):

PR #156, opened 2023-10-14
feat: allow disabling format on save per buffer

horriblename:
allows disabling format on save at runtime with `let b:formatsave = v:false`

### DIFF
--- a/docs/release-notes/rl-0.5.adoc
+++ b/docs/release-notes/rl-0.5.adoc
@@ -22,6 +22,8 @@ https://github.com/horriblename[horriblename]:
 
 * Add lua LSP and treesitter support, and neodev.nvim plugin support
 
+* Add <<opt-vim.lsp.mappings.toggleFormatOnSave>> keybind
+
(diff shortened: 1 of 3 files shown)


NotAShelf:
could be a good idea to have a (configurable) keybind instead of hoping the user remembers what the command is

NotAShelf:
the `l` key is usually reserved for LSP actions, not sure if it's the correct group.

Perhaps we can add a keybind such as `<leader>s` for saving and something else like `<leader>sd` for save-disable?

horriblename:
it is an LSP thing tho is it not? since formatOnSave is defined in `vim.lsp.formatOnSave`

NotAShelf:
it is, in all technicality, but to an end user formatting is something the editor should do on a whim

which is why I think it should be a single top level key

NotAShelf:
though, lets not re-invent the wheel for now. I think the way things are right now are fine

NotAShelf (approved):

PR #158, opened 2023-10-16
utility/surround: Add mappings for nvim-surround

ksonj:
The default mappings for nvim-leap and nvim-surround conflict (i.e. nvim-surround uses `S` in visual mode). This change adds options to adapt the mappings for nvim-surround directly from the surround-module.

### DIFF
--- a/docs/release-notes/rl-0.5.adoc
+++ b/docs/release-notes/rl-0.5.adoc
@@ -69,3 +69,7 @@ https://github.com/jacekpoz[jacekpoz]:
 * Updated clangd to 16
 
 * Disabled `useSystemClipboard` by default
+
+https://github.com/ksonj[ksonj]:
+
+* Add support to change mappings to utility/surround
--- a/modules/utility/surround/config.nix
+++ b/modules/utility/surround/config.nix
@@ -6,14 +6,38 @@
 with lib;
 with builtins; let
   cfg = config.vim.utility.surround;
+  self = import ./surround.nix {inherit lib;};
+  mappingDefinitions = self.options.vim.utility.surround.mappings;
+  mappings = addDescriptionsToMappings cfg.mappings mappingDefinitions;
 in {
-  config = mkIf (cfg.enable) {
-    vim.startPlugins = [
-      "nvim-surround"
-    ];
+  config = mkIf cfg.enable {
+    vim = {
+      startPlugins = [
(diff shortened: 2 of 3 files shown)


NotAShelf:
Looks good! Could you update the release notes for v0.5 before I merge this?

NotAShelf (dismissed):
LGTM

NotAShelf:
Looks like the option definitions are conflicting, and the files are missing a quick `nix fmt`

ksonj:
Ah, nice about the checks. Not sure what to do about the conflict, though. I guess it surfaces now that we've made the defaultx explicit, but the conflict existed before. Should we change the default mappings for surround then?

ksonj:
or make it conditional on whether leap is enabled?

NotAShelf:
Changing default mappings for surround sounds like the sensible option, high chance user will want to have them both enabled and still retain configurability.

ksonj:
There is some discussion as to which defaults would be a sensible choice at https://github.com/ggandor/leap.nvim/discussions/59 

Personally I map all surround commands to the "gz" variants as proposed in https://github.com/ggandor/leap.nvim/discussions/59#discussioncomment-3943598. But I'm not sure whether one wants to inflict such a radical deviation from the upstream defaults upon users of this flake?

So the minimally invasive change would to only remap the visual mode `S` mapping. But now whatever we come up with (`gS`?, `Z`?, `gz`?, ...) is most certainly inconsistent with the other mappings of surround.

Alternatively one might leave the defaults at `null`, only actually touch `vim.maps` if a mapping is explicitly set and let the default conflict persist invisibly as before? Not optimal either.

hm, after having written this down, I actually think I prefer choosing the more or less consistent set of gz-mappings and add an explanation to the description of the `enable` option.

what do you think?

ksonj:
I wasn't certain again whether to set the individual mapping defaults to the values from upstream, `null` or to our vendored defaults. I discarded the first because of the possibility of a change in the upstream defaults and then settled on the latter since that gives a nicer default UX (as if you stay with the defaults, the actual values are documented in the appendix). Setting `useVendoredKeybindings` to `false` now sets all keybindings to `null`, thus accepting the implicit conflict of visual mode `S` when using nvim-leap.

NotAShelf:
Thanks!

Issue #133, opened 2023-09-07
extraConfig to pass into builtin plugin's `setup()`


horriblename:
### ⚠️ Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find a similar feature request

### 🏷️ Feature Type

API Additions

### 🔖 Feature description

I'd like to add custom options that should be passed into the `setup` function of builtin plugins, 

### ✔️ Solution

I have an idea using a nix to lua transpiler.

Example:

```nix
# user's configuration
vim.filetree.nvimTree = {
   enable = true;
   openOnSetup = false;
   # this attrset will be transpiled into lua:
   extraConfig = {
      select_prompts = false;
      view = { width = 30; };
      renderer.special_files = ["Cargo.toml" "Makefile"];
      on_attach = { _type = "_rawLua", code = ''
         function(bufnr)
            vim.keymap.set("n", "x", ":DoSomething<CR>", {buffer=bufnr})
         end
      ''};
   };
};
```

should generate this lua code (except uglier):

```lua
require("nvim-tree").setup({
   open_on_setup = true,
   select_prompts = false,
   view = { width = 30 },
   renderer = {special_files = {"Cargo.toml", "Makefile"}},
   on_attach = function(bufnr)
      vim.keymap.set("n", "x", ":DoSomething<CR>", {buffer=bufnr})
   end
})
```

We'll have to rewrite existing options in terms of `extraConfig` instead:

```nix
options.vim.filetree.nvimTree = {
      enable = mkEnableOption "...";
      openOnSetup = mkOption { ... };
      extraConfig = mkOption {
         type = types.anything;
      };
};

# cfg = config.vim.filetree.nvimTree
config.extraConfig = {
   open_on_setup = cfg.open_on_setup
}

config.vim.luaConfigRC.nvimtreelua = nvim.dag.entryAnywhere ''
   require 'nvim-tree'.setup(${convertToLua cfg.extraConfig})
''
```

## The transpiler

with some minor improvements we can use `lib.nvim.lua.expToLua`

~~The transpiler should convert a nix value into a lua value, the only restriction is lambdas aren't allowed.~~

~~I have a working implementation [here](https://github.com/horriblename/dots.nix/blob/master/modules/home/terminal/nvim/lib/toLua.nix)~~

[...]

PR #159, opened 2023-10-20
languages/python: add black-and-isort formatter

ksonj:
Adds two additional formatters for python:

- isort, to only sort imports
- black-and-isort, which runs black + isort in a compatible manner


NotAShelf:
Tested and seems to be working great. Thanks!

NotAShelf (approved):

NotAShelf:
Requested a review from @horriblename for second opinion, I don't work with Python too often.

horriblename (approved):

NotAShelf:
> LGTM, although we should probably start considering adding proper support for multiple formatters, and also LSP/linters

yeah the LSP/linter system is in a desperate need of a rework, might be discussed in the same narrative as #133 after the release of 0.5, which should be soon

PR #160, opened 2023-10-20
statusline/lualine: disable LSP status inside toggleterm buffer

NotAShelf:
Disables LSP status inside the "toggleterm" buffer by returning "" if current buftype is toggleterm.

### DIFF
--- a/modules/statusline/lualine/lualine.nix
+++ b/modules/statusline/lualine/lualine.nix
@@ -185,19 +185,29 @@ in {
           {
             {
               -- Lsp server name .
+
               function()
-                local msg = 'No Active Lsp'
                 local buf_ft = vim.api.nvim_buf_get_option(0, 'filetype')
+
+                -- Check if the current buffer type is "toggleterm"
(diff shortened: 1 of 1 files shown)


modules/statusline/lualine/lualine.nix
```diff
@@ -185,19 +185,29 @@ in {
           {
             {
               -- Lsp server name .
+
               function()
-                local msg = 'No Active Lsp'
                 local buf_ft = vim.api.nvim_buf_get_option(0, 'filetype')
```
FrothyMarrow, line 190: `nvim_buf_get_option()` is deprecated. Use `nvim_get_option_value()` instead.
```suggestion
                local buf_ft = vim.api.nvim_get_option_value('filetype', {})
```

FrothyMarrow (changes requested):

modules/statusline/lualine/lualine.nix
NotAShelf: Good catch.

PR #167, opened 2023-10-21
plugins: Replace null-ls with none-ls

ksonj:
null-ls is unmaintained as of August 2023 (see
https://github.com/jose-elias-alvarez/null-ls.nvim/issues/1621). none-ls appears to be the accepted replacement fork.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -916,6 +916,22 @@
         "type": "github"
       }
     },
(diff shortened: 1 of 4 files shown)


NotAShelf (commented):
Ah I was waiting for a fork to pop up, its great that null-ls lives on.

PR #169, opened 2023-10-21
fix enable option descriptions

ksonj:
I noticed a few oddly phrased descriptions for enable options in the docs.

mkEnableOption already adds the phrase "Whether to enable ..." to the beginning of the option description, such that the string argument should only be "thing to be enabled"

Also, sorry for the PR spam :grimacing:


NotAShelf:
It is very much appreciated, thank you!

PR #171, opened 2023-10-22
lsp: make key bindings configurable

ksonj:
This exposes the keybindings for the lsp commands to the configuration and maps a few more commands (e.g. `vim.lsp.buf.format()`) It also adds a description to be picked up by whichkey to make the whichkey UX a bit nicer (ie. we're now showing "Go to definition" instead of just "vim.lsp.buf.definition()")


NotAShelf:
looks good!

NotAShelf (approved):

horriblename (approved):
LGTM

PR #173, opened 2023-10-23
statusline/lualine: extensible sections

ksonj:
I have a usecase where I need to add an additional source to a section in the statusline. Since repeating and then extending the current defaults isn't very convenient, I've changed the type of `activeSection`s and `inactiveSection`s attributes to `listOf str`, added `extraActiveSection` and `extraInactiveSection` of the same type and then create a lua table from the concatentation of both settings in the config.

I immediately see two issues with this:

1. I am breaking existing configurations that customize the lualine sections
2. Having a nix-list of lua-values is a bit odd. Eg. adding a string would look like ``["'some_string'"]` :shrug: 

I'm not sure if (1.) is a problem. You're on 0.x still so I assume it's justifiable, but I'm of course willing to adapt this PR if you prefer a proper deprecation process or to keep the existing types.

For (2.) I think this somewhat is the nature of Nix but again I am willing to explore alternatives.

### DIFF
--- a/modules/statusline/lualine/config.nix
+++ b/modules/statusline/lualine/config.nix
@@ -5,6 +5,7 @@
 }:
 with lib; let
   cfg = config.vim.statusline.lualine;
+  luaTable = items: ''{${builtins.concatStringsSep "," items}}'';
 in {
   config = (mkIf cfg.enable) {
     vim.startPlugins = [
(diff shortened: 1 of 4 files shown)


modules/statusline/lualine/config.nix
```diff
@@ -5,6 +5,7 @@
 }:
 with lib; let
   cfg = config.vim.statusline.lualine;
+  luaTable = items: ''{${builtins.concatStringsSep "," items}}'';
```
NotAShelf, line 8: I'd prefer if helper functions were in the extended library. This can go with test of the lua helpers.

modules/statusline/lualine/config.nix
```diff
@@ -32,21 +33,21 @@ in {
...
-          lualine_c = ${cfg.inactiveSection.c},
-          lualine_x = ${cfg.inactiveSection.x},
-          lualine_y = ${cfg.inactiveSection.y},
-          lualine_z = ${cfg.inactiveSection.z},
+          lualine_a = ${luaTable (cfg.inactiveSection.a ++ cfg.extraInactiveSection.a)},
+          lualine_b = ${luaTable (cfg.inactiveSection.b ++ cfg.extraInactiveSection.b)},
+          lualine_c = ${luaTable (cfg.inactiveSection.c ++ cfg.extraInactiveSection.c)},
+          lualine_x = ${luaTable (cfg.inactiveSection.x ++ cfg.extraInactiveSection.x)},
+          lualine_y = ${luaTable (cfg.inactiveSection.y ++ cfg.extraInactiveSection.y)},
+          lualine_z = ${luaTable (cfg.inactiveSection.z ++ cfg.extraInactiveSection.z)},
```
NotAShelf, line 50: I think this is an interesting take. Why don't we instead concat all strings passed to `activeSection...` attribute before appending the extra section? The groundwork seems to be ready, in a sense that the type of sections changed from string to a list of string.

Would be good for "futureproofing" and allow lualine module to separate lualine components in their own list element.

modules/statusline/lualine/lualine.nix
```diff
@@ -276,47 +286,111 @@ in {
...
-          }
-        '';
+            }
+          ''
+        ];
+      };
+    };
+    extraActiveSection = {
+      a = mkOption {
+        type = types.listOf types.str;
```
NotAShelf, line 296: A minor nitpick, but I'd rather use `with types; listOf str` in option definitions.

NotAShelf:
> I immediately see two issues with this:
> 
>     1. I am breaking existing configurations that customize the lualine sections
> 
>     2. Having a nix-list of lua-values is a bit odd. Eg. adding a string would look like ``["'some_string'"]` 🤷
> 
> 
> I'm not sure if (1.) is a problem. You're on 0.x still so I assume it's justifiable, but I'm of course willing to adapt this PR if you prefer a proper deprecation process or to keep the existing types.
> 
> For (2.) I think this somewhat is the nature of Nix but again I am willing to explore alternatives.

(1.) is not really a problem. Breaking changes to be expected, such is the nature of improvement. I'll make a pinned issue describing the changes.

For (2.), it really is difficult to convert Nix to Lua in a way that it *looks* pretty. We already have some lua helpers, on top of the one in nixpkgs, and I am yet to see a converter that can make generated lua look good. We can think about the conversion method a little more, but as long as it's *functional* we should be good.

NotAShelf (changes requested):
Looks good. Minor nitpick regarding the location of the function and type definitions, should be good afterwards.

modules/statusline/lualine/config.nix
ksonj: :+1:  there even is a listToLuaTable function already, though it's not exactly the right type here

NotAShelf (approved):

FrothyMarrow (approved):

PR #179, opened 2023-11-07
treewide: move from 'with lib' to 'inherit (lib) ...'

FrothyMarrow:
Changes most modules, only ones remaining are `lsp` and `languages` (I hope). The opportunity is yours @NotAShelf.


FrothyMarrow:
`nix flake check` is also failing on the `assert-assertions`  branch because of the same reason.

PR #185, opened 2023-12-07
fix: disable empty winbar when breadcrumbs are disabled

horriblename:
Fixes bug where an empty winbar is shown even when breadcrumbs are disabled


NotAShelf (approved):
I was so sure I've tested this

LGTM

PR #187, opened 2023-12-09
docs: refactor

NotAShelf:
Completely switches off optionDocBook and follow home-manager documentation formatting for rendering.

Some sections may be incomplete as a result of the asciidoc -> markdown conversion. Do point out if you notice any.


NotAShelf:
Seems like the latest version of nmd pads codeblocks on the top for some reason. I'll need to look into that.

PR #198, opened 2024-01-18
Language/markdown: Added LSP and fixed Glow not working

Donnerinoern:
# Description

I don't know if it's good practice to combine two commits into one PR like this, but they both touch on the same module, so I figured it would probably be okay. Correct me if I'm wrong.

The first commit adds marksman as a markdown LSP. It also provides the same options that the other language modules do.

The second commit fixes Glow, which I noticed wasn't working correctly (or at all). As with the first commit, new option(s) are provided. There is now an option for changing the preview keybind, which by default is the same bind that it was before (\<Leader>p).

## Type of change

- Bug fix (I guess. Glow not working hasn't been reported as an issue, though)
- New feature

## Checklist

- [x] My code follows the style and contributing guidelines of this project.
- [x] I ran Alejandra to format my code (`nix fmt`).
- [x] I have performed a self-review of my own code and tested it.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [x] My changes generate no new warnings.
- [ ] This change requires a documentation update.
- [ ] I have updated the documentation accordingly.

## Screenshots & Logs

The maximal config completely fails to build for me, so I replaced it with my own. When using my own, I get no new errors, nor any new warnings.
image
image
image

### DIFF
--- a/docs/release-notes/rl-0.6.md
+++ b/docs/release-notes/rl-0.6.md
@@ -27,3 +27,5 @@ Release notes for release 0.6
 - Added Gruvbox theme
 
 - Added marksman LSP for Markdown
+
+- Fixed Markdown-previewer Glow not working and added an option for changing the preview keybind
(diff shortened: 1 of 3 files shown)


NotAShelf (approved):
LGTM

FrothyMarrow (approved):
👍

NotAShelf:
Thanks!

PR #200, opened 2024-01-24
languages: add css and tailwind language support

NotAShelf:
Adds support for CSS and Tailwindcss.

From what I can tell, Tailwindcss likely depends on css treesitter, but I do not want those two modules to be codependent on each other.

### DIFF
--- a/modules/languages/css.nix
+++ b/modules/languages/css.nix
@@ -1,0 +1,74 @@
+{
+  config,
+  pkgs,
+  lib,
+  ...
+}: let
+  inherit (builtins) attrNames;
+  inherit (lib) mkEnableOption mkOption mkIf mkMerge isList types nvim;
+
+  cfg = config.vim.languages.css;
+
+  defaultServer = "vscode-langservers-extracted";
+  servers = {
+    vscode-langservers-extracted = {
+      package = pkgs.nodePackages.vscode-langservers-extracted;
+      lspConfig = ''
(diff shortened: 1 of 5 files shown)


modules/languages/css.nix
```diff
@@ -0,0 +1,72 @@
...
+  cfg = config.vim.languages.css;
+
+  defaultServer = "vscode-langservers-extracted";
+  servers = {
+    vscode-langservers-extracted = {
+      package = pkgs.nodePackages.vscode-langservers-extracted;
+      lspConfig = ''
+        -- enable (broadcasting) snippet capability for completion
+        -- see <https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#cssls>
+        local capabilities = vim.lsp.protocol.make_client_capabilities()
```
horriblename, line 19: this might override the original definition of `capabilities`, I'm not sure where its defined but maybe configure that one globally (if it doesn't affect other LSPs), or rename this one to `css_capabilities`

horriblename (changes requested):

modules/languages/css.nix
NotAShelf: that's a good call

horriblename:
you forgot to rename the one on line 20 lol

horriblename:
LGTM

 can't approve on mobile so I'll do it tomorrow

horriblename (approved):

PR #202, opened 2024-01-31
Fix java lsp not launching

Amanse:
java lsp pacakge in nixpkgs changed the binary name, using getExe fix this situation


NotAShelf (approved):
I was hoping I would not forget about the change, seems I did anyway.

LGTM, thanks.

PR #205, opened 2024-02-07
languages/nim: init

NotAShelf:
Take two, properly rebased this time.

- Implements Nim LSP and formatter support which I have tested briefly. Seems to be working, but LSP sometimes takes a second or two to kick in before we actually get diagnostics


FrothyMarrow:
Tested the PR on Linux and it seems to work fine. On MacOS, NimLSP crashes because of segment violation. I have created an issue (https://github.com/PMunch/nimlsp/issues/178#issue-2128106096), meanwhile maybe mark the support for nim as broken on MacOS?

NotAShelf:
I'll add an assertion

NotAShelf:
@FrothyMarrow pushed a commit to assert on Darwin systems. Given I don't have a way to test it, would appreciate if you could test it.

FrothyMarrow:
The assert works but it will also fail the maximal builds on Darwin. I believe it would be better to have it either disabled by default or add a check there as well.

FrothyMarrow (approved):
👍

PR #212, opened 2024-02-11
filetree/nvimtree: fix default systemOpen on darwin

FrothyMarrow:
Makes the default `systemOpen` option platform specific.

### DIFF
--- a/modules/filetree/nvimtree/nvimtree.nix
+++ b/modules/filetree/nvimtree/nvimtree.nix
@@ -191,7 +191,12 @@ in {
       };
 
       cmd = mkOption {
-        default = "${pkgs.xdg-utils}/bin/xdg-open";
+        default =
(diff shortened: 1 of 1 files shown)


NotAShelf (commented):
maybe we can add another if here that checks if the system is stdenv.isLinux and throws otherwise?

FrothyMarrow:
I hadn't realized that the systems could now be overridden. I've added the check for Linux. 👍

NotAShelf:
I'm tempted to merge, but "couldn't find an utility to open lol" doesn't really help the user. Thinking maybe there could be instructions (i.e. "change this option to fix this") to avoid softlocks?

NotAShelf (approved):
looks bad to me, but I will show mercy

FrothyMarrow:
Soft

PR #217, opened 2024-02-12
language/markdown: added markdown-previewer.nvim

Donnerinoern:
# Description

Added markdown-previewer.nvim as a Markdown previewer.

## Type of change

- New feature

## Checklist

- [x] My code follows the style and contributing guidelines of this project.
- [x] I ran Alejandra to format my code (`nix fmt`).
- [x] I have performed a self-review of my own code and tested it.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [x] My changes generate no new warnings.
- [ ] This change requires a documentation update.
- [ ] I have updated the documentation accordingly.

## Screenshots & Logs

image

PR #220, opened 2024-02-15
docs: add missing documentation entry; define renamed options

NotAShelf:
Adds the missing documentation entry for #217, and ensures the changes are backwards compatible through the usage of `mkRenamedOptionModule` to warn the users that have the old option still defined instead of erroring out.


NotAShelf:
CC @Donnerinoern for a quick review, should be ready to merge if no objections.

Donnerinoern (approved):
LGTM.

PR #224, opened 2024-02-17
utility/images: add image.nvim

NotAShelf:


FrothyMarrow (commented):
Looks good to me, although it seems that `ueberzugpp` is broken on macOS. Tried it both on wezterm and alacritty. Add an assert for now, I will try to look into it.

FrothyMarrow (approved):

PR #226, opened 2024-02-17
Fix lua null conversion

horriblename:
fixes two bugs with null conversions:

1. `expToLua` converts null to null (should be nil)
2. `toLuaObject` converts null to ""

### DIFF
--- a/lib/lua.nix
+++ b/lib/lua.nix
@@ -1,7 +1,7 @@
 # Helpers for converting values to lua
 {lib}: let
   inherit (lib) mapAttrsToList filterAttrs concatStringsSep concatMapStringsSep stringToCharacters boolToString;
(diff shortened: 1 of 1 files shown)


NotAShelf (dismissed):
LGTM! But just a thought, could there be any occasion that the conversion of null to `""` is desirable?

horriblename:
I don't think so, it would cause runtime errors as that might result in invalid lua syntax.

in fact I think "" shouldn't be there at all and should just error out

I'll read through the usage to see how its used

horriblename:
everywhere toLuaObject is used "" is not valid syntax, so I changed it to throw an error

NotAShelf (approved):
LGTM, thanks!

Issue #221, opened 2024-02-16
Plugin mappings have a higher "description" priority than manually-set mappings.


Nowaaru:
### ⚠️ Please verify that this bug has NOT been reported before.

- [X] I checked all existing issues and didn't find a similar issue

### Description

I've noticed that mappings that have an "implicit" `desc` field end up completely overtaking my manual mappings. Instead of having `[plugin-name]` as the description, I would prefer that it be the fallback if there's no mapping made with the same binding whilst also possessing a `desc` field.

### 👟 Reproduction steps

I have a keybind with "`<Nop>`" as the action:
```nix
{
	"<leader>f" = {
		action = "<Nop>";
		desc = "File";
	};
};
```
and two Telescope keybinds related to files:
```nix
{
	mappings = {
	    findFiles = mkDefault "<leader>ff"; 
        findProjects = mkDefault "<leader>fp"; 
	};
};
```
Because the keybind comes from a plugin, it completely disappears my File description. 

### 👀 Expected behavior

I expect that my File mapping would take over.

If not that, then optional DAG support would be pretty cool where mappings are made during the "pluginBinding" entry and user bindings are after "pluginBinding."

Also could just settle with `mkDefault` for plugin bindings and `mkOverride` for user bindings. :thinking:   

### 😓 Actual Behavior

My plugin mapping takes over and makes the keybind group the name of the Plugin (in this case, `+Telescope`) when the purpose of the group is supposed to be for general file things. 

### 💻 Metadata

 - system: `"x86_64-linux"`  - host os: `Linux 6.7.1, NixOS, 24.05 (Uakari), 24.05.20240210.d934204`  - multi-user?: `yes`  - sandbox: `yes`  - version: `nix-env (Nix) 2.18.1`  - channels(root): `"nixos, nixpkgs"`  - nixpkgs: `/home/noire/.nix-defexpr/channels/nixpkgs`

### 📝 Relevant log output

```bash
No logs.
```

NotAShelf:
`mkDefault` is just an alias for giving the option a lower priority. If you are setting those options in your own config, either don't set `mkDefault` or pass `mkForce` when you need to override high-priority defaults.

Nowaaru:
Yes, but even trying `mkOverride 0 [...]` and setting the option directly (which I think is the same priority as mkDefault?) doesn't solve the issue. `mkForce` doesn't help either.

FrothyMarrow:
[...]

PR #234, opened 2024-02-21
utility/binds/which-key: allow configurable category labels

### LINKED ISSUE #222: "Folder" option for vim.maps

- [X] I checked and didn't find a similar feature request

### 🏷️ Feature Type

API Additions

### 🔖 Feature description

Currently, it's not possible to write "description-only" mappings to the flake configuration. 
It expects an `action` field, but that would turn it into a key mapping. 

It's pretty odd to have a flood of `+prefix` with no way to change it, which kind-of defeats the purpose of which-key because I have no clue which-prefix I need to go into lol

### ✔️ Solution

I'd just prefer:
```nix
{
	"<leader><leader>" = {
		desc = "Miscellaneous";
	};
}
```
without Home Manager complaining about an expected `action` field.

### ❓ Alternatives

Tried using `<Nop>`.

### 📝 Additional Context

Tangentially related to #221.

FrothyMarrow:
Fixes #222, maybe #221? I did not really understand the issue there, hoping this is it.

### DIFF
--- a/modules/utility/binds/which-key/config.nix
+++ b/modules/utility/binds/which-key/config.nix
@@ -3,14 +3,17 @@
   lib,
   ...
 }: let
-  inherit (lib) mkIf nvim;
+  inherit (lib.modules) mkIf;
+  inherit (lib.strings) optionalString;
+  inherit (lib.nvim.lua) toLuaObject;
+  inherit (lib.nvim.dag) entryAnywhere;
 
   cfg = config.vim.binds.whichKey;
 in {
   config = mkIf (cfg.enable) {
     vim.startPlugins = ["which-key"];
 
-    vim.luaConfigRC.whichkey = nvim.dag.entryAnywhere ''
+    vim.luaConfigRC.whichkey = entryAnywhere ''
       local wk = require("which-key")
       wk.setup ({
         key_labels = {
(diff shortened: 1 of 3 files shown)


NotAShelf (commented):
regiser :trollface:

FrothyMarrow:
Reran the job, worked this time. 🤔

NotAShelf:
Probably got ratelimited by github.

modules/utility/binds/which-key/config.nix
```diff
@@ -4,6 +4,7 @@
   ...
 }: let
   inherit (lib) mkIf nvim;
```
NotAShelf, line 6: In the `restructure` branch, I'm making lib calls more explicit. Would make my life easier if this PR also converted to the new format.

```nix
inherit (lib.options) mkIf
inherit (lib.nvim.lua) toLuaObject
```

modules/utility/binds/which-key/which-key.nix
```diff
@@ -1,7 +1,77 @@
-{lib, ...}: let
-  inherit (lib) mkEnableOption;
+{
+  config,
+  lib,
+  ...
+}: let
+  inherit (lib) mkEnableOption mkOption mkOptionType types;
```
NotAShelf, line 6: Same as my previous comment. See https://noogle.dev for parent attributes of each lib call.

modules/utility/binds/which-key/which-key.nix
```diff
@@ -1,7 +1,77 @@
...
+
+  whichKeyCategory = mkOptionType {
+    name = "whichKeyCategory";
+    description = "The label to use for the which-key keybind helper menu.";
+    check = label:
+      if label == null && !(lib.isAttr label)
```
NotAShelf, line 12: To my knowledge, null comparisons (foo == null) are a no-op in nix. Use builtins.isNull

modules/utility/binds/which-key/which-key.nix
```diff
@@ -1,7 +1,77 @@
...
+      then true
+      else false;
+  };
 in {
   options.vim.binds.whichKey = {
     enable = mkEnableOption "which-key keybind helper menu";
+
+    register = mkOption {
+      description = "Register label for which-key keybind helper menu";
+      type = types.attrsOf whichKeyCategory;
```
NotAShelf, line 24: Move types to a top level inherit

modules/utility/binds/which-key/which-key.nix
```diff
@@ -1,7 +1,77 @@
...
+      default =
```
NotAShelf, line 25: a mkMerge would be better here

```nix

mkMerge [
	(mkIf config.vim.tabline.nvimBufferline.enable {
		"<leader>b" = {name = "+Buffer";};
  	"<leader>bm" = {name = "BufferLineMove";};
    "<leader>bs" = {name = "BufferLineSort";};
    "<leader>bsi" = {name = "BufferLineSortById";};
   }) 
]

File: modules/utility/binds/which-key/which-key.nix (hunk 3 above)
Comment (NotAShelf) [reply modules/utility/binds/which-key/which-key.nix:12]: I misremembered, nvm.

File: modules/utility/binds/which-key/which-key.nix (hunk 5 above)
Comment (FrothyMarrow) [reply modules/utility/binds/which-key/which-key.nix:25]: The reason I preferred using attrset and `//` operator is because the check function in mkOptionType expects:
```nix
{ name = "+Telescope"; }
```
instead of:
```nix
{ "<leader>f" = { name = "+Telescope"; }; }
```
Which simplifies the validation because it is not trying to extract arbitrary named elements from the anonymous attrset. Resolving this for now and leaving this for the future. 🤞

modules/utility/binds/which-key/which-key.nix
FrothyMarrow: The opposite is true. 😕

modules/utility/binds/which-key/which-key.nix
eclairevoyant: Surely using an existing option type would make more sense than rolling your own here? Then you don't need to write your own `check` function.

I would normally use `attrsOf submodule` here.

FrothyMarrow: Indeed, I was just unaware about the existence of submodules. Thank you 👍

FrothyMarrow:
Rethought the idea, going to close this PR and open a new one.

PR #235, opened 2024-02-25
utility/binds/which-key: allow configurable category labels

FrothyMarrow:
Continuing #234


modules/filetree/nvimtree/config.nix
```diff
@@ -4,7 +4,7 @@
   pkgs,
   ...
 }: let
-  inherit (lib) mkIf mkMerge mkBinding nvim boolToString;
+  inherit (lib) mkIf mkMerge mkBinding nvim boolToString defaultAttributes;
```
NotAShelf, line 7: make imports explicit

lib/stdlib-extended.nix
```diff
@@ -68,6 +68,8 @@ in
     mkSetLuaBinding = binding: action:
       mkLuaBinding binding.value action binding.description;
 
+    defaultAttributes = attr: self.mapAttrs (name: value: self.mkDefault value) attr;
```
NotAShelf, line 71: I'm not sure if I like `defaultAttributes` as the function name. It should hint towards either the fact that it's a label (e.g. defaultLabels) or the fact that it's a constructor function (e.g mkDefaultLabelAttr)

NotAShelf (changes requested):

NotAShelf (approved):

horriblename (approved):
LGTM

PR #237, opened 2024-02-29
wrapper: migrate to `makeNeovimUnstable` wrapper

FrothyMarrow:
Directly wraps neovim using `wrapNeovimUnstable` instead of wrapping it through the `legacyWrapper` under `wrapNeovim`. 

I took a quick look at its commit history and it seems pretty stable despite its name. Regardless, I would appreciate a second opinion on this.


NotAShelf:
I have no doubt in its stability, seeing how actively it is used.

Looks good, I'll provide a review later today.

PR #241, opened 2024-03-24
treewide: make lib calls explicit

FrothyMarrow:
Should make almost all of the lib calls explicit. 

Renames the  function parameters of `diagnostics` option from `diagnostics` and `defaultDiagnostics` to `diagnosticsProviders` and `defaultDiagnosticsProvider` respectively, to avoid naming conflicts.


NotAShelf (approved):
LGTM

PR #242, opened 2024-03-24
assistant/chatgpt: Add jackMort/ChatGPT.nvim

ksonj:
This adds https://github.com/jackMort/ChatGPT.nvim to the assistant module and enables configuration of keybinds. There's a lot more stuff that could be configured, but I guess those can be exposed once someone finds the need.

### DIFF
--- a/configuration.nix
+++ b/configuration.nix
@@ -221,6 +221,7 @@ inputs: let
           enable = isMaximal;
           cmp.enable = isMaximal;
         };
+        chatgpt.enable = isMaximal;
       };
 
       vim.session = {
--- a/docs/release-notes/rl-0.6.md
+++ b/docs/release-notes/rl-0.6.md
@@ -48,3 +48,7 @@ Release notes for release 0.6
 [elijahimmer](https://github.com/elijahimmer)
 
 - Added rose-pine theme
+
+[frothymarrow](https://github.com/frothymarrow)
+
+- Added option `vim.luaPackages` to wrap neovim with extra Lua packages.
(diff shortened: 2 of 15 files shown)


NotAShelf (dismissed):
This looks good. Seeing it doesn't add any options, it should be compatible with the setupOpts changes of v0.6

ksonj:
Nice, didn't see that option yet. I'll add that then

ksonj:
wait, are we moving towards `mkPluginSetupOption` or away from it? I see it's not currently used in `main`?

NotAShelf:
We're moving towards `mkPluginSetupOption`, see https://github.com/NotAShelf/neovim-flake/pull/238 for large-scale examples.

It's currently not used in main as it's somewhat of a breaking change, v0.6 will feature it.

ksonj:
And the idea is to keep the options organized in a way that translates 1:1 to what upstream uses, such that you can call `toLuaObject` on the attrset to directly pass it to `setup()`?

NotAShelf:
Not much to do with upstream, but it allows you to create a lua config from a structured nix attribute set instead of abusing string interpolation with each option.

NotAShelf:
Sorry for the delay, I've been holding this PR off to avoid potential conflicts with the v0.6 branch. Would you mind if I rebase and merge this to v0.6?

ksonj:
Absolutely, please do

NotAShelf:
god the merge conflicts

I'll resolve them manually later today

NotAShelf:
Continued in #258

PR #245, opened 2024-04-07
treewide: restructure neovim-flake tree

NotAShelf:
**Take three of #197 and #225.**

---

## DONE
1. Reduces dependency cycles and layered imports in `modules/`
   - All plugins have been moved to `modules/plugins/`
   - Option definitions separated from configurations
   - `Basic` and `Core` have been isolated as `modules/core` and `modules/neovim` (previously basic)`
     - `core` will contain wrapper and build related files
     - `neovim` will contain the most basic vim/neovim options that are passed to base configurations.

2. Cleans up and re-organizes lib so that files and functions are easier to search by name.
3. Moves vim-dirtytalk to plugins.

## TODO/TBD

Things I want to do:

- Move neovim config builder/helper functions into lib, and call from lib where necessary. This has been an issue while implementing the home manager module, and I'd like to avoid repeating the same mistake again.

Things I might want to do:

- Get rid of inputs per-plugin. It's a nice idea, but flakes suck. Hundreds of inputs that a user will *likely* not even think about overriding in their whole usage of this flake is, I assume, burdening. If my fellow co-maintainers @FrothyMarrow and @horriblename agree; I'd like to fall back to either customizable `package` options, or a packages list that gets passed to Neovim wrapper.
- Other suggestions by contributors and maintainers alike


NotAShelf:
I officially consider this ready to go, it's not perfect but I will revise the structure as the need arises. I wanted to revise lib a bit, but I think that's going to be to big of an endeavor right now - one that I do not have the energy for. Unless @horriblename or @FrothyMarrow have a request or any requests for changes, this can be merged.

PR #250, opened 2024-04-14
completion/nvim-cmp: add `vim.autocomplete.autoComplete`

jacekpoz:
works on my machine :+1:

### DIFF
--- a/modules/plugins/completion/nvim-cmp/nvim-cmp.nix
+++ b/modules/plugins/completion/nvim-cmp/nvim-cmp.nix
@@ -1,11 +1,17 @@
 {lib, ...}: let
   inherit (lib.options) mkEnableOption mkOption literalMD;
   inherit (lib.nvim.binds) mkMappingOption;
-  inherit (lib.types) enum attrsOf nullOr str;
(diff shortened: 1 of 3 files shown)


modules/plugins/completion/nvim-cmp/nvim-cmp.nix
```diff
@@ -7,6 +7,8 @@ in {
     autocomplete = {
       enable = mkEnableOption "enable autocomplete" // {default = false;};
 
+      alwaysComplete = mkEnableOption "Automatically show completion." // {default = true;};
```
NotAShelf, line 10: `mkEnableOption` automatically prefixes the description with "Whether to enable". This should be `mkEnableOption "automatically showing completion menu."` so that the documentation is grammatically correct.

While you're on that, maybe remove the "enable" prefix from the option definition above.

NotAShelf (changes requested):
Minor change needed, LGTM otherwise.

jacekpoz:
tested this one, works as before

NotAShelf (approved):
LGTM, thanks!

PR #255, opened 2024-04-19
Fix rebase mistakes

horriblename:
I might have made a mistake or two during rebase. This hopefully un-undoes all refactors

### DIFF
--- a/modules/plugins/assistant/copilot/copilot.nix
+++ b/modules/plugins/assistant/copilot/copilot.nix
@@ -4,7 +4,7 @@
   lib,
   ...
 }: let
-  inherit (lib) mkRenamedOptionModule;
+  inherit (lib.modules) mkRenamedOptionModule;
   inherit (lib.options) mkEnableOption mkOption;
   inherit (lib.types) nullOr str enum float;
   inherit (lib.nvim.types) mkPluginSetupOption;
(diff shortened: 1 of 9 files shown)


modules/plugins/filetree/nvimtree/nvimtree.nix
```diff
@@ -3,11 +3,87 @@
...
+    tab = "tab";
+    notify = "notify";
+    ui = "ui";
+  };
+
+  renamedSetupOpts = flatten (genSetupOptRenames [] migrationTable);
+
+  # Note: I cut a few corners so it only works in this specific case
+  # if the parent of a nested option needs to be renamed, this would not work
+  genSetupOptRenames = path: table:
```
NotAShelf, line 74: Maybe this should be moved to the lib? I understand it may not be suitable for all renamed option situations, but I'm not very keen on tossing utility functions into random components. 

Perhaps lib/plugins/nvimtree.nix and `lib.nvim.plugins.renamedSetupOpts` would do the trickç

NotAShelf (approved):

modules/plugins/filetree/nvimtree/nvimtree.nix
horriblename: man that commit was not supposed to be in this branch

I'll remove it from this PR, but I'll update #254 with your suggestions later

PR #259, opened 2024-04-20
treewide: post-restructure refactoring

NotAShelf:
Cleans up a few bad decisions from the latest restructure, adds documentation to some options and makes spellchecking more versatile.


deviantsemicolon:
This PR is either good or bad, and you should either feel good or feel bad about it. Idk I haven't read it.

PR #261, opened 2024-04-23
plugins/treesitter: migrate legacy codebase; refactor

NotAShelf:
- Cleans up treesitter plugin directory to account for future treesitter extension plugins.
- Adds an internal `defaultGrammars` option to be added to treesitter grammars if the user chooses to have them added.
- Moves `treesitter-context` to the new setupOpts API.
- Override vim's `syntax on` with Treesitter while `vim.treesitter.highlight` is set to true.


FrothyMarrow:
We should cleanup the language modules of duplicate treesitter entires that are now included by default.

NotAShelf:
> We should cleanup the language modules of duplicate treesitter entires that are now included by default.

what entries are they?

PR #271, opened 2024-04-28
languages/typst: add typst language support

FrothyMarrow:
Adds support for typst with `typst-lsp` LSP and `typstfmt` as the default formatter.

Contributor: ItsSorae


NotAShelf:
This is missing the changelog entry. Would be appreciated if you could initiate and populate v0.7 docs (as the next release) in a future PR.

PR #277, opened 2024-05-01
modules: clean up plugins directory

NotAShelf:
- Move border configuration from `plugins` to `neovim/global`
- Make icons configurable. The icons can be written to a lua table at the beginning of the lua config to make it accessible within lua files.
- Make `vim.diagnostics` configurable


NotAShelf:
Needs to be approached again from the ground up.

PR #288, opened 2024-05-08
wrapper/rc: add vim.extraLuaFiles

NotAShelf:
* Adds `vim.additionalLuafiles`
  Allows the user to specify a list of lua files that will be called via `luafile`. All paths
  that are passed to this option are checked by `builtins.isPath` so attempting to source paths
  that do not exist do not result in a broken Lua configuration.

* Bumps flake inputs to their latest version (might fix nil timing out on large directories)

### DIFF
--- a/modules/wrapper/rc/config.nix
+++ b/modules/wrapper/rc/config.nix
@@ -133,6 +133,14 @@ in {
       configRC = {
         globalsScript = entryAnywhere (concatStringsSep "\n" globalsScript);
 
+        # Call additional lua files with :luafile in Vimscript
+        # section of the configuration, only after
+        # the luaScript section  has been evaluated
+        extraLuaFiles = let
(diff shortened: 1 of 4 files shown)


modules/wrapper/rc/config.nix
```diff
@@ -133,6 +133,21 @@ in {
...
         globalsScript = entryAnywhere (concatStringsSep "\n" globalsScript);
 
+        # Call additional lua files with :luafile in Vimscript
+        # section of the configuration, only after
+        # the luaScript section  has been evaluated
+        extraLuaFiles = let
+          callLuaFiles =
+            map (
+              file:
+                if builtins.isPath file
```
FrothyMarrow, line 143: `builtins.isPath` doesn't check if a file exists, it only checks if the type is `path`. 
If it is a path, it must be valid otherwise it will cause an evaluation error before it reaches this code.
Paths that are `str` will always throw for the same reason above.

PR #289, opened 2024-05-12
lib: do not filter out empty attrs in toLuaObject

horriblename:
there is a single use of `__empty` in the codebase, I need to double check what it's doing before I remove it

### DIFF
--- a/lib/lua.nix
+++ b/lib/lua.nix
@@ -3,7 +3,7 @@
   inherit (builtins) hasAttr head throw typeOf isList isAttrs isBool isInt isString isPath isFloat toJSON;
   inherit (lib.attrsets) mapAttrsToList filterAttrs;
   inherit (lib.strings) concatStringsSep concatMapStringsSep stringToCharacters;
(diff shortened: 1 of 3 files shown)


NotAShelf (dismissed):
LGTM

NotAShelf:
can be merged once @FrothyMarrow also approves

FrothyMarrow (commented):
Add a warning for the users in case they are using `__empty`.

lib/lua.nix
```diff
@@ -66,7 +66,10 @@ in rec {
       if isLuaInline args
       then args.expr
       else if hasAttr "__empty" args
-      then "{ }"
+      then
+        lib.warn ''
```
NotAShelf, line 70: Inherit `warn` from (lib.trivial) on top. We do not use aliases.

NotAShelf (changes requested):
minor nit, LGTM otherwise

NotAShelf (approved):

PR #292, opened 2024-05-15
add ocaml support

jacekpoz:
this adds support for the ocaml lsp and formatter

everything seems to be working, the only issue is the lsp gives me a warning about not being able to find the formatter on format:
image
but it still formats so ¯\\\_(ツ)\_/¯

I think it could be fixed by adding `ocamlformat` to nvf's environment but I don't know how to do that sowwy uwu

### DIFF
--- a/configuration.nix
+++ b/configuration.nix
@@ -73,6 +73,7 @@ inputs: let
         go.enable = isMaximal;
         elixir.enable = isMaximal;
         zig.enable = isMaximal;
+        ocaml.enable = isMaximal;
         python.enable = isMaximal;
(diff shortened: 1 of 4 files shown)


NotAShelf:
> I think it could be fixed by adding ocamlformat to nvf's environment but I don't know how to do that sowwy uwu

nvf exposes a `vim.extraPackages` option to insert packages into the runtime PATH. You can consider adding ocamlformat there.

jacekpoz:
that worked thanks

NotAShelf (approved):
LGTM, thanks

PR #303, opened 2024-06-11
which-key: allow null in registers

diniamo:
Useful for disabling registers defined by nvf. Although that shouldn't be needed in the first place, since defining static registers doesn't account for the user changing the keymaps. For example, if I change all my telescope binds to start with `<leader>s` instead of `<leader>f, then I'll have both prefixes appear in which-key.

See https://github.com/diniamo/niqs/blob/72039800e859e4f790c3a0d8a7759c06d9c222b9/home/nvf/plugins/telescope.nix

### DIFF
--- a/modules/plugins/utility/binds/which-key/which-key.nix
+++ b/modules/plugins/utility/binds/which-key/which-key.nix
@@ -4,14 +4,14 @@
   ...
 }: let
   inherit (lib.options) mkEnableOption mkOption;
-  inherit (lib.types) attrsOf str;
+  inherit (lib.types) attrsOf nullOr str;
 in {
   options.vim.binds.whichKey = {
     enable = mkEnableOption "which-key keybind helper menu";
 
     register = mkOption {
       description = "Register label for which-key keybind helper menu";
-      type = attrsOf str;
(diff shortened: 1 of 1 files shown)


FrothyMarrow:
There isn't a seamless way to integrate whick-key and the alternatives have these problems:

- Defaults remove all the predefined labels.
- Having an option per module will only allow labels for existing modules.

Hence, an overridable system was the compromise. It is similar to predefined mappings which also need to be set to `null` to remove them.

In your use case, which-key should not be used for registering descriptions for Telescope mappings. It can parse descriptions directly from the mappings, and this is an issue with the Telescope module itself. I will make a PR to add those descriptions. At most, you should need to `null` out the `+Telescope` category label.

We can consider flipping the attribute set around and having "+Telescope" as the overridable but I am not sure if more people are interested in remapping or renaming the label itself.

FrothyMarrow (approved):

NotAShelf (approved):

NotAShelf:
binds are honestly a whole can of works that I'd rather not open right now

that doesn't hold this PR, however, so merging unless no objections arise

PR #306, opened 2024-06-14
modules/theme: fix dag location

diniamo:


NotAShelf (changes requested):
Looks good overall.

But since this is partially a breaking change, you should document this change in the release notes. It also breaks any DAG placement in user configurations depending on `"theme"` but asserting that would reserve the name `"theme"` in unrelated cases.

NotAShelf (approved):
I don't like having to introduce multiple lua blocks, but this seems like the best solution so far.

PR #316, opened 2024-06-25
modules/nvim-cmp: add edge case for autopairs in cr mapping

diniamo:
Tested


NotAShelf (approved):
The diff looks good, not able to test until later today but I trust your testing.

PR #317, opened 2024-06-25
modules/autopairs: use boolToString instead of toLuaObject

diniamo:
Not that important, but I didn't realize that function existed before


NotAShelf:
toLuaObject should automatically apply the correct type (which in this case, is a string). Not sure if this change is *really* necessary. But it is probably a more concise solution.

NotAShelf (approved):

PR #325, opened 2024-07-08
modules: switch to gerg's neovim-wrapper

NotAShelf:


NotAShelf:
self-merge gang rise up

PR #327, opened 2024-07-10
Feature: basic lz.n setup

horriblename:
things maybe to do:
- integrate into extraPlugins instead of using a separate option

todo:
- vomit code into init.lua when lazy.enable = false
- doc: add lazy section in keymaps
- doc: recommend `vim.lazy.plugins` instead

bug:
- ~~there's this stupid bug where if you give only one plugin it doesn't work, I'll fix it upstream later~~ fixed


horriblename:
moved to #407

PR #333, opened 2024-07-13
treewide: make the entire generated config lua based

diniamo:


horriblename (changes requested):
LGTM, aside from the missing `local`

horriblename:
> `Requires formatting: ./modules/neovim/init/basic.nix`

and fix formatting pls

diniamo:
Thanks

modules/wrapper/rc/options.nix
```diff
@@ -3,13 +3,19 @@
...
   inherit (lib.strings) optionalString;
-  inherit (lib.types) str oneOf attrs lines listOf either path bool;
+  inherit (lib.types) str attrs lines listOf either path bool;
   inherit (lib.nvim.types) dagOf;
   inherit (lib.nvim.lua) listToLuaTable;
+
   cfg = config.vim;
 in {
+  imports = [
+    (mkRemovedOptionModule ["vim" "configRC"] "Migrate your configRC sections to lua, and add them to luaConfigRC instead.")
```
NotAShelf, line 16: It would be nice to add
1. A "migration guide" to the top of the changelog (like we did for setupOpts in 0.6)
2. Mention said guide here

modules/wrapper/rc/options.nix
```diff
@@ -120,35 +126,21 @@ in {
...
-
-      example = literalMD ''
-        ```vim
-        " Set the tab size to 4 spaces
-        set tabstop=4
-        set shiftwidth=4
-        set expandtab
-        ```
-      '';
+      description = "The internal DAG used to configure plugins";
```
NotAShelf, line 143: Since this is internal, we might want to hide pluginRC from the docs. In which case, pass `visible = false;` as well.

Unless we want people to see it?

NotAShelf (commented):
LGTM overall, I'll provide an approving review after you address my previous comments and I test this on my own config.

modules/wrapper/rc/options.nix
diniamo: Up to you, I'm against leaving it visible, since you can always just do `entryAfter ["pluginConfigs"]` and it will end up being a same thing, except without having to look up the plugin's entry name.

diniamo: But `builtConfigRC` already doesn't appear in the manual, isn't `visible = false` implied?

NotAShelf (approved):

PR #346, opened 2024-07-20
modules: add helper scripts

NotAShelf:
- **modules: add helpers to display init.lua and its store path**
- **docs: update 0.7 release notes**

### DIFF
--- a/modules/default.nix
+++ b/modules/default.nix
@@ -105,10 +105,30 @@ inputs: {
     inherit (vimOptions) viAlias vimAlias withRuby withNodeJs withPython3;
     inherit extraLuaPackages extraPython3Packages;
   };
+
+  # Additional helper scripts for printing and displaying nvf configuration
+  # in your commandline.
+  printConfig = pkgs.writeShellScriptBin "print-nvf-config" ''
+    cat << EOF
+      ${vimOptions.builtLuaConfigRC}
+    EOF
+  '';
+
+  printConfigPath = pkgs.writeShellScriptBin "print-nvf-config-path" ''
+    realpath ${pkgs.writeTextFile {
+      name = "nvf-init.lua";
+      text = vimOptions.builtLuaConfigRC;
+    }}
+  '';
 in {
   inherit (module) options config;
   inherit (module._module.args) pkgs;
 
-  # expose wrapped neovim-package
-  neovim = neovim-wrapped;
+  # Expose wrapped neovim-package for userspace
+  # or module consumption.
+  neovim = pkgs.symlinkJoin {
+    name = "nvf-with-helpers";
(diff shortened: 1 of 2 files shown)


modules/default.nix
```diff
@@ -105,10 +105,30 @@ inputs: {
     inherit (vimOptions) viAlias vimAlias withRuby withNodeJs withPython3;
     inherit extraLuaPackages extraPython3Packages;
   };
+
+  # Additional helper scripts for printing and displaying nvf configuration
+  # in your commandline.
+  printConfig = pkgs.writeShellScriptBin "print-nvf-config" ''
+    cat << EOF
+      ${vimOptions.builtLuaConfigRC}
+    EOF
```
diniamo, line 114: Could call the other script, and cat the result of that to save some space.

modules/default.nix
```diff
@@ -105,10 +105,30 @@ inputs: {
...
 in {
   inherit (module) options config;
   inherit (module._module.args) pkgs;
 
-  # expose wrapped neovim-package
-  neovim = neovim-wrapped;
+  # Expose wrapped neovim-package for userspace
+  # or module consumption.
+  neovim = pkgs.symlinkJoin {
+    name = "Neovim";
```
diniamo, line 130: Make this lower-case, as it's a derivation name.

diniamo (changes requested):

modules/default.nix
```diff
@@ -105,10 +105,30 @@ inputs: {
...
+
+  # Additional helper scripts for printing and displaying nvf configuration
+  # in your commandline.
+  printConfig = pkgs.writeShellScriptBin "print-nvf-config" ''
+    cat << EOF
+      ${vimOptions.builtLuaConfigRC}
+    EOF
+  '';
+
+  printConfigPath = pkgs.writeShellScriptBin "print-nvf-config-path" ''
```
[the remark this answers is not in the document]
NotAShelf: Why?

modules/default.nix
NotAShelf: I don't think that saves any space.

diniamo: Well this way, 2 files contain the built config, if we cat the first file instead of including the whole thing again, then only 1 file will.

modules/default.nix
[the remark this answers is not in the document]
NotAShelf: It's faster that's why.

modules/default.nix
NotAShelf: A derivation calling another derivation requires using runCommand - *not* a method that I prefer unless I have to.

Those scripts amount to roughly 55kbs in total, so imo the size difference is negligible.

diniamo (approved):

PR #355, opened 2024-08-12
v0.7

NotAShelf:
## Branchout for 0.7

- Until this PR is merged, any new features must target the `v0.7` branch to minimize merge conflicts.
- Breaking changes must be disclosed, and remedied: if something broke, it should be documented.
- Here be dragons! Expect unstable behaviour until this is merged, and especially if the PR is a draft.

## Significant Changes

For changes not documented here, see the changelog. This section only mentions large, breaking changes that are worth noting. For changes not mentioned here, please see the change-log.

1. Maps rewrite (#352 and #360)
2. Lazy-loading support (#407)
3. Base16 theming (#389)
4. Completion rewrite, Luasnip (#398)
5. New Language Modules
   - Kotlin support (#390)
   - Scala support (#399)
   - Neorg support (#413)
   - C# support (#428)
   - Julia support (#431)
   - Vala support (#432)
   - ASM support (#447) 
6. Rewrite of Visual modules (#442) 
   - Some options have been renamed, `vim.visuals.enable` removed
7. All plugin inputs have been bumped to their absolute latest revisions.
8. Documentation improvements, cleanup (#440)
9. Deprecation of `vim.leaderKey` and similar `vim.*` options in favor of a migration to `vim.options` and `vim.globals` (#455)

PR #359, opened 2024-08-17
add language HCL

ppenguin:
Terraform doesn't register hcl and doesn't offer good DX if manually set for editing e.g. nomad HCL files.

My first nvf contribution. Maybe somebody could show how to integrate a formatter as well. Other than that it's already an improvement for editing non-Terraform files (e.g. `nomad`), but it could be probably better. (I vaguely remember the vscode plugin offered some more features).


NotAShelf:
Looks reasonable. Though, I will need you to format via `alejandra` and not `nixfmt`. The formatting check is failing.

Other than that, if there is a known formatter for HCL I can work on getting it added tomorrow. We will need null-ls (or none-ls) to support HCL for proper formatting.

ppenguin:
> Looks reasonable. 

Thanks for looking into it so quickly...

> Though, I will need you to format via `alejandra` and not `nixfmt`. The formatting check is failing.

Ah yes, different than `nixpkgs` :sweat_smile: 

While we're at it, is there a trick that I can (e.g. via `direnv`) use the active HM config for `nvf` but override certain attributes (like the formatter) in a devshell (when I'm in the `nvf` repo)? Or would I need to outsource my "standard" `nvf` config to a separate flake for that?

NotAShelf:
Looking at none-ls builtins, I see an [formatter source](https://github.com/nvimtools/none-ls.nvim/blob/main/doc/BUILTINS.md#terraform_fmt for Terraform. This could be used to add formatter support via `terraform fmt`. Refer to other language modules on how formatter submodules are added - the process is extremely straightforward but I'll be around if you have questions.

> While we're at it, is there a trick that I can (e.g. via direnv) use the active HM config for nvf but override certain attributes (like the formatter) in a devshell (when I'm in the nvf repo)? Or would I need to outsource my "standard" nvf config to a separate flake for that?

Afraid there isn't a way to impurely override configuration per-project, but we do provide a formatter that will format the code for you with `nix fmt`. I will consider making this a pre-commit hook, or at least have pre-commit warn you before you commit the changes.

ppenguin:
> Looking at none-ls builtins, I see an [formatter source](https://github.com/nvimtools/none-ls.nvim/blob/main/doc/BUILTINS.md#terraform_fmt) for Terraform. This could be used to add formatter support via `terraform fmt`. Refer to other language modules on how formatter submodules are added - the process is extremely straightforward but I'll be around if you have questions.
> 

I finally got arround to finish this one. In hindsight I might have gone a bit overboard with (unnecessary) extensibility (inspired by the `bash` setup), but given the fact that `terraform-ls` is a bit too generic for `nomad`, there may be more specific solutions in the funture that we might then easily integrate.

modules/plugins/languages/hcl.nix
```diff
@@ -0,0 +1,116 @@
...
+      nullConfig = ''
+        table.insert(
+          ls_sources,
+          null_ls.builtins.formatting.hclfmt.with({
+            command = "${lib.getExe package}",
+          })
+        )
+      '';
+    };
+  };
```
NotAShelf, line 42: Please avoid `¶ec`. It is a bad pattern, and I'd like to avoid it altogether in nvf if possible. You can refer to the lsp package with `cfg.lsp.package`

modules/plugins/languages/hcl.nix
```diff
@@ -0,0 +1,116 @@
...
+      type = mkOption {
+        description = "HCL formatter to use";
+        type = enum (attrNames formats);
+        default = defaultFormat;
+      };
+      package = mkOption {
+        description = "HCL formatter package";
+        type = package;
+        default = formats.${cfg.format.type}.package;
+      };
```
NotAShelf, line 77: Could you move description to the bottom? This is mostly a personal nit, but `description` is the option with most potential to span over multiple lines so it's better to have it at the bottom of `mkOption` args.

modules/plugins/languages/hcl.nix
```diff
@@ -0,0 +1,116 @@
...
+            local bo = vim.bo[opts.buf]
+            bo.tabstop = 2
+            bo.shiftwidth = 2
+            bo.softtabstop = 2
+          end
+        })
+
+        local ft = require('Comment.ft')
+        ft
+          .set('hcl', '#%s')
```
NotAShelf, line 96: Is this an official recommendation or a personal preference? If the latter, please consider adding a toggle and using `optionalString`

NotAShelf (changes requested):
Sorry for the delay, GitHub never gave me any notifications about your comment.

Left a few comments, but nothing too big.

NotAShelf:
Sorry for the delay in my review, I'm very busy with IRL stuff at the moment.

We're currently at the final stretch for merging the next big release (see #355) and I've frozen the main branch to avoid merge conflicts. If you could rebase this around the `v0.7` branch, I should be able to provide a final review or/and merge this PR tomorrow.

ppenguin:
Thanks

> If you could rebase this around the `v0.7` branch, I should be able to provide a final review or/and merge this PR tomorrow.

Done (just pushed ther rebased one)

NotAShelf (approved):
LGTM, but there seems to be an issue with the merge. Please pulling `v0.7` with a rebase. There is a conflict between main (which is pulled) and `v0.7` which you are targeting.

Also, a changelog entry wouldn't go amiss :)

ppenguin:
Ah, the good old _rebase_ vs _merge_ mess :wink: 
Should be ok now...

NotAShelf:
All goor, merging once CI passes.

PR #360, opened 2024-08-18
maps: allow same key on multiple mode

horriblename:
A second rewrite of keymaps cuz the last one doesn't allow different action on different modes

### DIFF
--- a/modules/neovim/mappings/config.nix
+++ b/modules/neovim/mappings/config.nix
@@ -3,32 +3,75 @@
   lib,
   ...
 }: let
-  inherit (lib.modules) mkIf;
+  inherit (lib.modules) mkIf mkMerge;
+  inherit (lib.trivial) pipe;
+  inherit (lib.attrsets) mapAttrsToList;
+  inherit (lib.lists) flatten;
+
+  legacyMapModes = {
+    normal = ["n"];
+    insert = ["i"];
+    select = ["s"];
+    visual = ["v"];
+    terminal = ["t"];
+    normalVisualOp = ["n" "v" "o"];
+    visualOnly = ["n" "x"];
+    operator = ["o"];
+    insertCommand = ["i" "c"];
+    lang = ["l"];
+    command = ["c"];
+  };
 
   cfg = config.vim;
 in {
   config = {
-    vim.maps = mkIf cfg.disableArrows {
-      "<up>" = {
-        mode = ["n" "i"];
-        action = "<nop>";
-        noremap = false;
-      };
-      "<down>" = {
-        mode = ["n" "i"];
-        action = "<nop>";
-        noremap = false;
-      };
-      "<left>" = {
-        mode = ["n" "i"];
-        action = "<nop>";
-        noremap = false;
-      };
-      "<right>" = {
-        mode = ["n" "i"];
-        action = "<nop>";
-        noremap = false;
-      };
-    };
+    vim.keymaps = mkMerge [
+      (
+        mkIf cfg.disableArrows [
+          {
+            key = "<up>";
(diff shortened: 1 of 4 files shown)


modules/neovim/mappings/options.nix
```diff
@@ -31,59 +31,72 @@
     options =
       mapConfigOptions
       // {
+        key = mkOption {
+          type = str;
+          description = ''
+            Key that triggers this keybind.
```
diniamo, line 37: `The key` rather, you can also make this string one line with `""`

modules/neovim/mappings/options.nix
```diff
@@ -31,59 +31,72 @@
...
+          '';
+        };
         mode = mkOption {
           type = either str (listOf str);
           description = ''
             The short-name of the mode to set the keymapping for. Passing an empty string is the equivalent of `:map`.
 
             See `:help map-modes` for a list of modes.
           '';
+          example = ''"nvc" for normal, visual and command mode'';
```
diniamo, line 47: nvc isn't a real short name from looking at the docs?

modules/neovim/mappings/options.nix
```diff
@@ -31,59 +31,72 @@
...
         };
       };
   };
 
   # legacy stuff
-  mapOption = submodule {
+  legacyMapOption = submodule {
     options = mapConfigOptions;
   };
```
diniamo, line 55: Redundant comment, you can also just inline this

modules/neovim/mappings/options.nix
```diff
@@ -31,59 +31,72 @@
...
 
   mapOptions = mode:
```
diniamo, line 57: Maybe name _this_ legacyMapOptions?

modules/wrapper/rc/config.nix
```diff
@@ -41,40 +41,11 @@ in {
...
-      // mapAttrs (_: legacyMap: legacyMap // {mode = namedModes.normalVisualOp;}) cfg.maps.normalVisualOp
-      // mapAttrs (_: legacyMap: legacyMap // {mode = namedModes.visualOnly;}) cfg.maps.visualOnly
-      // mapAttrs (_: legacyMap: legacyMap // {mode = namedModes.operator;}) cfg.maps.operator
-      // mapAttrs (_: legacyMap: legacyMap // {mode = namedModes.insertCommand;}) cfg.maps.insertCommand
-      // mapAttrs (_: legacyMap: legacyMap // {mode = namedModes.lang;}) cfg.maps.lang
-      // mapAttrs (_: legacyMap: legacyMap // {mode = namedModes.command;}) cfg.maps.command;
+    maps = concatLines (map toLuaKeymap cfg.keymaps);
 
-    keymaps = concatLines (map toLuaKeymap (attrsToList (filterAttrs (_: value: value != null) maps)));
+    keymaps = maps;
```
diniamo, line 48: Why??

modules/neovim/mappings/config.nix
```diff
@@ -3,32 +3,75 @@
...
+                  // {
+                    inherit key;
+                    mode = legacyMapModes.${oldMode};
+                  }
+              )
+              keybinds
+          ))
+          flatten
+        ]
+      )
```
diniamo, line 74: I actually quite like this compared to whatever we had before

diniamo (changes requested):

modules/neovim/mappings/options.nix
horriblename: oof, I thought nvim supports this, turns out no

diniamo (approved):
Should be good

PR #369, opened 2024-08-31
treewide: remove the theme section from the DAG order dependance

diniamo:
This is needed, since the DAG order breaks for users who don't use the builtin theming options.


jacekpoz:
tested, launches with no errors

jacekpoz (approved):

NotAShelf (approved):

PR #372, opened 2024-09-13
languages/python: add basedpyright & python-lsp-server

NotAShelf:
New python LSPs, rejoice Python users!


diniamo (approved):
LGTM besides random changes, they don't really matter though.

PR #378, opened 2024-09-22
R: adding LSP

Soliprem:
Adding LSP support for R. 
Can't get treesitter to work, but am including what I have so far (commenting out the activation) so someone else might give it a look. I must be missing something


NotAShelf:
Everything LGTM. I'm not sure why treesitter grammar would not work (but I can make a guess basede on the state of treesitter packaging.)

Are there any errors when you enable the treesitter grammar? Or does it simply not highlight anything?

NotAShelf:
Ah, I see the issue.

You should define treesitter.package with `mkGrammarOption`:

```nix
treesitter = {
      enable = mkEnableOption "R treesitter" // {default = config.vim.languages.enableTreesitter;};
      package = mkGrammarOption pkgs "r";
};
```

Soliprem:
Testing it locally gives me the same error:

Soliprem:
image

Soliprem:
Mind you, highlighting works, but I suspect it might be falling back onto the native syntax highlighting

NotAShelf:
Does maximal configuration work?

Soliprem:
I'm testing the binary generated with nix build .#maximal

NotAShelf:
Okay I ran into it as well. My best bet would be that either
a. the grammar is outdated (trying with a newer nixpkgs revision now)
or
b. the grammar is just faulty

Soliprem:
in case of a, we can just wait for a new nixpkgs version, in case of b I guess we just take out treesitter from the module?

NotAShelf:
We could get rid of the treesitter package and opt in for the LSP only. Appears that it's *not* the nixpkgs revision, unless it's an issue with TS packaging.

Soliprem:
huh, might want to report it on the nixpkgs repo as a bug with the package, then. In the meantime, I'd simply edit the PR to only include the LSP. Coming within a few minutes.

Soliprem:
is there any indication that beyond the error Treesitter was working? I didn't really know how to check, beyond running checkhealth, which didn't return anything interesting

NotAShelf:
My best guess is that the TS grammar is outdated. I'll try overriding the source and see if that is the case.

NotAShelf:
The *built* grammar seems to work. Not what I hoped would be the solution, but we can easily resolve the TS issue by using a built grammar it seems...

Soliprem:
Implementation-wise how does that work?

Soliprem:
wasn't that the content of the original PR? Using ```pkgs.vimPlugins.nvim-treesitter.builtGrammars.r``` ?

NotAShelf (changes requested):
LGTM, but needs changelog entry added before I can merge this.

NotAShelf (approved):

NotAShelf:
Merging once CI passes.

NotAShelf:
thank you :)

Issue #239, opened 2024-03-09
add documentations for vim.languages.*.extraDiagnostics 


rice-cracker-dev:
### ⚠️ Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find a similar feature request

### 🏷️ Feature Type

Other

### 🔖 Feature description

vim.languages.*.extraDiagnostics isn't very obvious to config

### ✔️ Solution

add better docs

NotAShelf:
Being worked on in #440

PR #382, opened 2024-09-26
treewide: language refactor

NotAShelf:
*Very* breaking change that moves formatter and diagnostic configurations to their standalone plugin/module counterparts for extended modularity. The new idea is that formatters/diagnostics are provided as module options, which are enabled by any language modules that need it. This allows for users to opt out of specific language modules, but still provide appropriate support for their desired languages.

This should improve configurability, compatibility and performance that were all impacted negatively by null-ls.

1. Adds `conform.nvim` and `nvim-lint` plugins for formatting and linting respectively.
   - [ ] Document adding formatters and linters (#239)
2. Deprecates null-ls, and removes any configuration that depends on null-ls (WIP)
3. Streamlines addition of new LSPs/formatters and so on.


LilleAila:
Could the plugin support for `conform.nvim` and `nvim-lint` be merged, and then language defaults and replacing null-ls be done in consecutive pull requests, keeping these plugins as optional for now?

NotAShelf:
Yes, that is a good idea. Something held back nvim-lint last time, but I can't remember what it was. I'll be revisiting this (+new LSP requests) when I get my network situation sorted.

horriblename (changes requested):

PR #399, opened 2024-10-05
languages/scala: Add scala language support

ksonj:
Scala language support

# Description

Adds basic support for Scala via nvim-metals.

This is very bare bones as of yet. It works, but metals can be configured quite extensively and none of those options are exposed via the module here yet.

## Type of change

- New feature (non-breaking change which adds functionality)

## Checklist

Please try to check at least a majority of the checklist before opening your pull request. Exceptions to this will be reviewed on a case by case basis.

- [x] My code follows the style and contributing guidelines of this project.
- [x] I ran Alejandra to format my code (`nix fmt`).
- [x] I have performed a self-review of my own code and tested it.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [x] My changes generate no new warnings.
- [ ] This change requires a documentation update.
- [ ] I have updated the documentation accordingly.
```

### CODE REVIEW

File: modules/plugins/languages/scala.nix [hunk 1]
```diff
@@ -0,0 +1,141 @@
...
+  ...
+}: let
+  inherit (lib) types optionalString;
+  inherit (lib.modules) mkIf mkMerge;
+  inherit (lib.nvim.types) mkGrammarOption;
+  inherit (lib.nvim.binds) mkMappingOption;
+  inherit (lib.options) mkOption mkEnableOption;
+  inherit (lib.nvim.dag) entryAfter;
+
+  mkBinding = binding: "vim.api.nvim_buf_set_keymap(bufnr, 'n', '${binding.value}', '<cmd>lua ${binding.action}<CR>', {noremap=true, silent=true, desc='${binding.description}'})";
```
NotAShelf, line 14: Is there a reason we're defining `mkBİnding` here and not using one of the binding functions from lib?

modules/plugins/languages/scala.nix
```diff
@@ -0,0 +1,141 @@
...
+  usingDap = config.vim.debugger.nvim-dap.enable && cfg.dap.enable;
+  usingLualine = config.vim.statusline.lualine.enable;
+in {
+  options.vim.languages.scala = {
+    enable = mkEnableOption "Scala language support";
+    treesitter = {
+      enable = mkEnableOption "Scala treesitter" // {default = config.vim.languages.enableTreesitter;};
+      package = mkGrammarOption pkgs "scala";
+    };
+    lsp = {
```
NotAShelf, line 36:
```suggestion
    };
    
    lsp = {
```

Here and in other instances. Cramped code is not very readable.

modules/plugins/languages/scala.nix
```diff
@@ -0,0 +1,141 @@
...
+      enable = mkEnableOption "Scala LSP support (metals)" // {default = config.vim.languages.enableLSP;};
+      package = mkOption {
```
NotAShelf, line 38: You could use `mkPackageOption` here.

modules/plugins/languages/scala.nix
```diff
@@ -0,0 +1,141 @@
...
+            )
+
+
+            -- without doing this, autocommands that deal with filetypes prohibit messages from being shown
+            vim.opt_global.shortmess:remove("F")
+
+            vim.cmd([[augroup lsp]])
+            vim.cmd([[autocmd!]])
+            vim.cmd([[autocmd FileType java,scala,sbt lua require('metals').initialize_or_attach(metals_config)]])
+            vim.cmd([[augroup end]])
```
NotAShelf, line 135: Please avoid Vimscript. Neovim has a Lua API for autocmds.

modules/plugins/languages/scala.nix
```diff
@@ -0,0 +1,141 @@
...
+    mkMerge [
+      (mkIf cfg.treesitter.enable {
+        vim.treesitter.enable = true;
+        vim.treesitter.grammars = [cfg.treesitter.package];
+      })
+      (mkIf (cfg.lsp.enable || cfg.dap.enable) {
+        vim = {
+          startPlugins = ["nvim-metals"];
+          luaConfigRC.nvim-metals = entryAfter ["lsp-setup"] ''
+            -- Scala nvim-metals config
```
NotAShelf, line 92: Redundant comment, `luaConfigRC.nvim-metals` will prepend a header like `-- SECTION: nvim-metals`

modules/plugins/languages/scala.nix
```diff
@@ -0,0 +1,141 @@
...
+  config = mkIf cfg.enable (
+    mkMerge [
+      (mkIf cfg.treesitter.enable {
+        vim.treesitter.enable = true;
+        vim.treesitter.grammars = [cfg.treesitter.package];
+      })
+      (mkIf (cfg.lsp.enable || cfg.dap.enable) {
+        vim = {
+          startPlugins = ["nvim-metals"];
+          luaConfigRC.nvim-metals = entryAfter ["lsp-setup"] ''
```
NotAShelf, line 91: use `pluginRC` here.

NotAShelf (changes requested):

modules/plugins/languages/scala.nix
ksonj: Might be just ignorance on my part. lsp/config.nix does a similar thing to set buffer local mappings, which is what I looked at.

ksonj:
BTW this PR also has nvim-metals still pointing to a fork of mine, because with upstream I am getting a build error on generating helptags. I hope to get this merged https://github.com/scalameta/nvim-metals/pull/693 to fix that.

modules/plugins/languages/scala.nix
ksonj: I've inlined this into `attach_metals_keymaps` now

ksonj:
Switched back scalameta/nvim-metals for the plugin source

modules/plugins/languages/scala.nix
```diff
@@ -0,0 +1,138 @@
...
+      enable = mkEnableOption "Scala treesitter" // {default = config.vim.languages.enableTreesitter;};
+      package = mkGrammarOption pkgs "scala";
+    };
+
+    lsp = {
+      enable = mkEnableOption "Scala LSP support (metals)" // {default = config.vim.languages.enableLSP;};
+      package = mkPackageOption pkgs "metals" {
+        default = ["metals"];
+      };
+      extraMappings = {
```
NotAShelf, line 37: Please add gaps between module options, here and in other cases of multi-line attribute sets that are after each other.

Inline options are fine, such as `enable` and `package` options.

modules/plugins/languages/scala.nix
```diff
@@ -0,0 +1,138 @@
...
+
+    dap = {
+      enable = mkOption {
+        description = "Scala Debug Adapter support";
+        type = types.bool;
+        default = config.vim.languages.enableDAP;
+      };
+      config = mkOption {
+        description = "Lua configuration for dap";
+        type = types.str;
```
NotAShelf, line 50: You can use luaInline type here (from nvim.lib)

modules/plugins/languages/scala.nix
```diff
@@ -0,0 +1,138 @@
...
+            {
+              type = "scala",
+              request = "launch",
+              name = "Test Target",
+              metals = {
+                runType = "testTarget",
+              },
+            },
+          }
+        '';
```
NotAShelf, line 71: After using luaInline type, this can be inlined with `mkLuaInline` from lib.generators (nixpkgs lib).

```nix
default = (mkLuaInline ''
	...
');

File: modules/plugins/languages/scala.nix [hunk 10]
```diff
@@ -0,0 +1,138 @@
...
+            vim.opt_global.shortmess:remove("F")
+
+            local lsp_group = vim.api.nvim_create_augroup('lsp', { clear = true })
+
+            vim.api.nvim_create_autocmd('FileType', {
+                group = lsp_group,
+                pattern = {'java', 'scala', 'sbt'},
+                callback = function()
+                    require('metals').initialize_or_attach(metals_config)
+                end,
```
NotAShelf, line 131: This entire section feels really off to me. I can't tell you right off the bat what would be better (if that's possible) but we are hardcoding *a lot* of values here with no good way of overriding.


To remedy this, we could:

a. try defining config options in lua tables with a distinctive prefix (e.g. `__nvf_metals_config`) and explicitly instruct users to modify said table to override in Lua configurations
*or*

b. Define most of those under a specific set under `vim.languages.scala`, say, `languages.scala.lsp.extraOpts` which gets passed to `toLuaObject from extended library.

Pick your poison.

modules/plugins/languages/scala.nix
```diff
@@ -0,0 +1,138 @@
+{
+  config,
+  pkgs,
+  lib,
+  ...
+}: let
+  inherit (lib) types optionalString mkPackageOption;
```
NotAShelf, line 7: We don't do `inherit (lib) types`. Please inherit each type that you use

`optionalString` is from `lib.strings`, `mkPackageOption` is from `lib.options` which is already in the scope - can be moved after `mkEnableOption`.

modules/plugins/languages/scala.nix
```diff
@@ -0,0 +1,138 @@
...
+            metals_config.handlers["textDocument/publishDiagnostics"] = vim.lsp.with(
+              vim.lsp.diagnostic.on_publish_diagnostics, {
+                virtual_text = {
+                  prefix = '⚠',
+                }
+              }
+            )
+
+            -- without doing this, autocommands that deal with filetypes prohibit messages from being shown
+            vim.opt_global.shortmess:remove("F")
```
NotAShelf, line 122: This probably should be documented or made optional with a `true` default.

NotAShelf (changes requested):
A few things I'd like to see changed for the sake of consistency and for documentation.

In addition, please rebase this PR on the `v0.7` branch as main is currently in a feature freeze.

modules/plugins/languages/scala.nix
ksonj: I went for an `extraSettings` option that is set as `metals_config.settings`. Not sure if that solves all your concerns, though. I saw that there's `mkPluginSetupOption` and I was inclined to use it, but decided against as I am not technically passing the table to `setup`.

NotAShelf:
If you could add a changelog entry, I think ı can merge this later today.

NotAShelf (approved):
There are some more things I'd like to change, but I think I'll handle those myself in the #382 eventually.

Thanks.

PR #406, opened 2024-10-06
visuals: move all plugins to setupOpts

NotAShelf:
Move visual plugins to the (not so) new `setupOpts` and give each plugin its own module under the visual directory.

Contains a lot of removals and renames under `vim.visuals` for the sake of consistency. Most `camelCase` names have been named to `kebab-case` since that is the dominant naming convention.

Depends on #407


NotAShelf:
I ain't fixing all that

PR #408, opened 2024-10-06
lib: remove pointless type restriction

horriblename:
idk why I made setupOpts `attrsOf anything` instead of `anything`


NotAShelf (approved):
I'm going to hope you have tested this and nothing explodes.

PR #422, opened 2024-10-21
spell: cleanup; add extraSpellFiles for arbitrary spellwords

NotAShelf:


modules/neovim/init/spellcheck.nix
```diff
@@ -25,9 +28,45 @@ in {
         A list of languages that should be used for spellchecking.
 
         To add your own language files, you may place your `spell`
-        directory in either `~/.config/nvim` or the
+        directory in either {file}`~/.config/nvim` or the
```
diniamo, line 31: It should be `~/.config/nvf`

modules/neovim/init/spellcheck.nix
```diff
@@ -25,9 +28,45 @@ in {
...
         [additionalRuntimePaths](#opt-vim.additionalRuntimePaths)
-        directory provided by **nvf**.
+        directory provided by nvf.
+      '';
+    };
+
+    extraSpellFiles = mkOption {
```
diniamo, line 37: The name is a little misleading, I wouldn't include "file" in it

modules/neovim/init/spellcheck.nix
```diff
@@ -25,9 +28,45 @@ in {
...
+      type = attrsOf (listOf str);
+      default = {"en.utf-8" = ["nvf" "word_you_want_to_add"];};
```
diniamo, line 39: "word_you_want_to_add" should not be a default

modules/neovim/init/spellcheck.nix
```diff
@@ -58,18 +97,52 @@ in {
...
+            for spellfile in $spellfilesJoined; do
+              # Hacky way to ensure that the mangled extensions are omitted from the
+              # joined spellfiles. E.g.
+              local extension=".add"
+              local name=$(basename $spellfile "$extension")
+              echo "Compiling spellfile: $spellfile"
+              nvim --headless --clean \
+                --cmd "mkspell $out/after/spell/"$name".add.spl $spellfile" -Es -n
+
+              ls -lah $out/after/spell
```
diniamo, line 126: Debug line?

modules/neovim/init/spellcheck.nix
```diff
@@ -58,18 +97,52 @@ in {
...
+        spellfilesJoined = pkgs.symlinkJoin {
+          name = "nvf-spellfiles-joined";
+          paths = mapAttrsToList (name: value: pkgs.writeTextDir "spell/${name}.add" (concatStringsSep "\n" value)) cfg.extraSpellFiles;
+          postBuild = ''
+            echo "Spellfiles joined"
+          '';
+        };
+
+        compileJoinedSpellfiles =
+          pkgs.runCommand "nvf-compile-spellfiles" {
```
diniamo, line 111: runCommandLocal should be better here

diniamo (dismissed):

modules/neovim/init/spellcheck.nix
NotAShelf: This is a testing value, default will be an empty attrset.

modules/neovim/init/spellcheck.nix
NotAShelf: Yes, this is a draft PR.

modules/neovim/init/spellcheck.nix
NotAShelf: True.

modules/neovim/init/spellcheck.nix
NotAShelf: It should actually be $XDG_CONFIG_HOME/nvf. I had forgotten to apply this change, whoops.

modules/neovim/init/spellcheck.nix
NotAShelf: What's a better name? I was considering `extraSpellWords` but that's also misleading since we expect a key-value pair of `name = list`

diniamo: extraSpellWords is fine, maybe extraSpellLists/extraSpellWordLists, although the last one is too long

NotAShelf: I think `extraSpellWords` is the best approach here. `extrasSpellLists` is equally misleading, and the latter is simply too long. The name gives you a hint to its purpose, and the documentation shows you the usage.

modules/neovim/init/spellcheck.nix
```diff
@@ -58,18 +96,53 @@ in {
...
-        callback = function()
-          vim.opt_local.spell = false
-        end,
-      })
-    '';
+    vim = {
+      additionalRuntimePaths = let
+        spellfilesJoined = pkgs.symlinkJoin {
+          name = "nvf-spellfiles-joined";
+          paths = mapAttrsToList (name: value: pkgs.writeTextDir "spell/${name}.add" (concatStringsSep "\n" value)) cfg.extraSpellFiles;
```
diniamo: Could use concatLines instead

modules/neovim/init/spellcheck.nix
```diff
@@ -58,18 +96,53 @@ in {
...
+          postBuild = ''
+            echo "Spellfiles joined"
+          '';
```
diniamo, line 106: Is the echo really necessary?

Also, just do the compilation here instead of in a separate derivation.

diniamo (changes requested):

modules/neovim/init/spellcheck.nix
NotAShelf: Necessary? No. But I prefer displaying build phases as they happen.

diniamo (approved):

PR #450, opened 2024-11-10
wrapper: get meta from wrapped neovim package

diniamo:


NotAShelf:
CI fails because of missing formatting.

PR #457, opened 2024-11-10
lazy: Remove unintuitive keymap mode default

horriblename:


modules/wrapper/lazy/lazy.nix
```diff
@@ -36,7 +36,6 @@
       mode = mkOption {
         description = "Modes to bind in";
         type = either str (listOf str);
-        default = ["n" "x" "s" "o"];
       };
```
NotAShelf, line 39: Maybe make this an example instead.

Also, please move description below type and default.

horriblename:
took the chance to reorder the options in lazy.nix, and fixed some bugs (forgot to wrap lazy.plugins.*.load in a lua func)

diniamo (approved):
Diff looks good

PR #477, opened 2024-11-27
modules/nu: init

Soliprem:
Following up on #374 

---

### DIFF
--- a/configuration.nix
+++ b/configuration.nix
@@ -76,7 +76,7 @@ isMaximal: {
       csharp.enable = isMaximal;
       julia.enable = isMaximal;
       vala.enable = isMaximal;
-      nu.enable = isMaximal;
+      nu.enable = false;
     };
 
     visuals = {
--- a/configuration.nix
+++ b/configuration.nix
@@ -76,6 +76,7 @@ isMaximal: {
       csharp.enable = isMaximal;
       julia.enable = isMaximal;
       vala.enable = isMaximal;
+      nu.enable = isMaximal;
     };
 
     visuals = {
--- a/modules/plugins/languages/nu.nix
+++ b/modules/plugins/languages/nu.nix
@@ -42,9 +42,9 @@ in {
     lsp = {
       enable = mkEnableOption "Nu LSP support" // {default = config.vim.languages.enableLSP;};
       server = mkOption {
-        description = "Nu LSP server to use";
         type = str;
         default = defaultServer;
+        description = "Nu LSP server to use";
       };
 
       package = mkOption {
--- a/modules/plugins/languages/nu.nix
+++ b/modules/plugins/languages/nu.nix
@@ -71,10 +71,10 @@ in {
       };
 
       package = mkOption {
-        description = "Nu LSP server package, or the command to run as a list of strings";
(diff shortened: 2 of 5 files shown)


configuration.nix
```diff
@@ -76,6 +76,7 @@ isMaximal: {
       csharp.enable = isMaximal;
       julia.enable = isMaximal;
       vala.enable = isMaximal;
+      nu.enable = isMaximal;
```
NotAShelf, line 79: I don't think this needs to be enabled in the maximal module. We've been inflating its size for languages that are rarely used for a while now, I would like to avoid that.

modules/plugins/languages/nu.nix
```diff
@@ -0,0 +1,113 @@
...
+  #             command = "${cfg.format.package}/bin/nufmt",
+  #             args = { "--stdin" },
+  #             to_stdin = true
+  #           },
+  #           factory = null_helpers.formatter_factory
+  #         }
+  #       )
+  #     '';
+  #   };
+  # };
```
NotAShelf, line 53: Is this a TODO? Or does it not work at all?

modules/plugins/languages/nu.nix
```diff
@@ -0,0 +1,113 @@
...
+        description = "Nu LSP server to use";
+        type = str;
+        default = defaultServer;
+      };
+
+      package = mkOption {
+        description = "Nu LSP server package, or the command to run as a list of strings";
+        example = ''[(lib.getExe pkgs.nushell) "--lsp"]'';
+        type = either package (listOf str);
+        default = servers.${cfg.lsp.server}.package;
```
NotAShelf, line 77:
```suggestion
      package = mkOption {
        type = either package (listOf str);
        default = servers.${cfg.lsp.server}.package;
        example = ''[(lib.getExe pkgs.nushell) "--lsp"]'';
        description = "Nu LSP server package, or the command to run as a list of strings"; 
```

1. type
2. default
3. example
4. description

is the ideal ordering for mkOption args.

NotAShelf (changes requested):

modules/plugins/languages/nu.nix
Soliprem: it's a TODO, current comment version doesn't work

Soliprem: TODO as "left for future work, when nufmt has a more stable schema and gets added to null-ls"

modules/plugins/languages/nu.nix
```diff
@@ -0,0 +1,70 @@
...
+      enable = mkEnableOption "Nu treesitter" // {default = config.vim.languages.enableTreesitter;};
+      package = mkGrammarOption pkgs "nu";
+    };
+
+    lsp = {
+      enable = mkEnableOption "Nu LSP support" // {default = config.vim.languages.enableLSP;};
+      server = mkOption {
+        description = "Nu LSP server to use";
+        type = str;
+        default = defaultServer;
```
NotAShelf, line 47:
```suggestion
        type = str;
        default = defaultServer;
        description = "Nu LSP server to use";
```

PR #480, opened 2024-12-02
neovim: gradually deprecate shorthand `vim.*` aliases to `vim.options` and `vim.globals`

NotAShelf:
Slowly deprecates top-level options that used to set Neovim's `vim.opt` options in favor of `vim.options` and `vim.globals` module options.


modules/wrapper/rc/options.nix
```diff
@@ -206,6 +222,13 @@ in {
             default = true;
             description = "Enable word wrapping.";
           };
+
+          signcolumn = mkOption {
+            type = bool;
```
horriblename, line 227: should be `either bool str`

"auto" and "yes:3" aren't exactly booleans

modules/wrapper/rc/options.nix
```diff
@@ -211,6 +237,16 @@ in {
...
 
+          signcolumn = mkOption {
+            type = either str bool;
+            default = true;
+            apply = x:
+              if isBool x
+              then toVimBool x # convert to a yes/no str
+              else x;
+            description = "Show the sign column";
+          };
```
diniamo, line 248: Please just have it be a string, and remove the toVimBool function

modules/wrapper/rc/options.nix
```diff
@@ -273,7 +313,7 @@ in {
         if [](#opt-vim.enableLuaLoader) is set to true.
       '';
 
-      example = literalExpression ''"$${builtins.readFile ./my-lua-config-pre.lua}"'';
+      example = literalExpression ''''${builtins.readFile ./my-lua-config-pre.lua}'';
```
diniamo, line 316:
```suggestion
      example = literalExpression "\${builtins.readFile ./my-lua-config-pre.lua}";
```

modules/neovim/init/basic.nix
```diff
@@ -106,63 +95,55 @@ in {
...
+      })
+
+      (optionalAttrs (cfg.lineNumberMode == "number") {
+        number = true;
+      })
+
+      (optionalAttrs (cfg.lineNumberMode == "relNumber") {
+        number = true;
+        relativenumber = true;
+      })
```
diniamo, line 140: We should just deprecate these too, no point in abstractions

diniamo (changes requested):

modules/neovim/init/basic.nix
NotAShelf: I'm not keen on aggressive deprecations outside the 'release cycle.' We can consider fully deprecating those options in the 0.8 branchout.

PR #490, opened 2024-12-09
languages/typst: add live preview for tinymist

NotAShelf:
- **meta: update release info (0.8)**
- **languages/typst: add `typst-preview.nvim` as an extension**


Soliprem (approved):
LGTM. Tested with and without custom options, seems to work as expected.

PR #494, opened 2024-12-10
ci: build previws for the website on doc changes

NotAShelf:


.github/workflows/docs-preview.yml
```diff
@@ -0,0 +1,172 @@
...
+          git config --global user.name "GitHub Actions"
+          git config --global user.email "actions@github.com"
+
+          # Set the GitHub token for authentication
+          git remote set-url origin https://x-access-token:${{ secrets.GITHUB_TOKEN }}@github.com/${{ github.repository }}
+
+          # Add and commit the changes
+          git add --all
+          git commit -m "Deploy PR #${PR_NUMBER} preview" || echo "No changes to commit"
+          git push --force origin $BRANCH_NAME
```
horriblename, line 65: race condition when multiple PRs are updated

maybe a simple exponential retry is good enough for now

```
for i in 0..10 {
  git pull
  if git push then break; else sleep $randomtime; done
}
```

.github/workflows/docs-preview.yml
```diff
@@ -0,0 +1,172 @@
...
+          git config --global user.name "GitHub Actions"
+          git config --global user.email "actions@github.com"
+
+          # Set the GitHub token for authentication
+          git remote set-url origin https://x-access-token:${{ secrets.GITHUB_TOKEN }}@github.com/${{ github.repository }}
+
+          # Add and commit the changes (only if there's something to delete)
+          git add .
+          git diff --quiet || git commit -m "Remove preview for PR #${PR_NUMBER}"
+          git push origin $BRANCH_NAME
```
horriblename, line 161: same as above. race condition

horriblename (changes requested):

.github/workflows/docs-preview.yml
NotAShelf: Is this not solved by co-currency opts? Or did that affect only page deployments.

horriblename: me dum

yea that should be good enough, though why is --forced needed here?

NotAShelf: It did fix something, but I can't recall which. Might've been worth documenting, but a tad too late.

PR #500, opened 2024-12-20
wrapper/lazy: add beforeSetup option

diniamo:
Useful for avoiding a million require calls in setupOpts.


NotAShelf (approved):
LGTM

horriblename:
I think you should just not use setupModule and setupOpts at this point tbh

that said I'm not against this change

horriblename (approved):

PR #502, opened 2024-12-20
wrapper/lazy: allow luaInline as a type in keys

diniamo:
One way of using keys is `{ "<C-x>", mode = "n" }`, which isn't possible in nvf without luaInline


horriblename:
does `keys = [{key = "<C-x>"; mode = "n";}]` not work?

diniamo:
Nothing works I have no idea why

horriblename:
works on my machine :tm:

diniamo:
I mean yeah tbf I'm just closing this. I didn't know there was a way to do it without mixed tables

PR #508, opened 2024-12-21
languages/solidity: init solidity module

amadaluzia:
## What have I added?
This adds support for the Solidity language into nvf. LSP support is not on nixpkgs, nor DAP support but the best I could have done (treesitter support) is there.

This also fixes the Haskell LSP command, I have no idea why it didn't use the `--lsp` flag by default.

## Have I tested it?
Yes, treesitter works.

### DIFF
--- a/configuration.nix
+++ b/configuration.nix
@@ -74,6 +74,7 @@ isMaximal: {
       julia.enable = false;
       vala.enable = false;
       scala.enable = false;
+      solidity.enable = false;
       r.enable = false;
       gleam.enable = false;
       dart.enable = false;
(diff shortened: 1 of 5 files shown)


NotAShelf:
I'm not sure if this *really* warrants a module. nvf already provides a way of adding treesitter grammars, and an option just to add the grammar for a language I'm hearing of for the first time does not sound very reasonable.

I'm not going to close this PR, in case LSPs or debug adapters ever get into nixpkgs, but I'm also not looking to merge it as is. You may consider using an overlay (like we do for nil) to package both of those in nvf and submitting a complete module if Solidity support is important to you.

diniamo:
lol but I will

amadaluzia:
Haven't had time to make a new branch with the LSP patch, I may be able to get this done once I have my new monitor come in.

I'd be happy for someone taking over for me though! :^)

PR #510, opened 2024-12-23
completion/blink.cmp: add

horriblename:
added blink.cmp

- [x] idk if I should keep enableSharedCmpSources
- [x] ~~why do I get extra backslash sometimes when using lsp?~~ can't repro anymore, prolly fixed upstream
- [x] should I raise prio of lsp snippets (i think it's deprio'd now)
~~- [ ] update script~~ (used `allowBuiltinFetchGit` instead)
- [x] lsp capabilities seems to be wrong, ldp snippets not showing
- [x] cmp glue stuff
- [x] lspkind

### DIFF
--- a/docs/release-notes/rl-0.8.md
+++ b/docs/release-notes/rl-0.8.md
@@ -47,6 +47,12 @@
 
 - Add Haskell support under `vim.languages.haskell` using [haskell-tools.nvim].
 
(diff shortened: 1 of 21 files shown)


rseichter:
This is exactly the completion plugin I was hoping for. 👍

PR #511, opened 2024-12-23
languages/zig: disable lsp format-on-save

diniamo:


NotAShelf:
Could you add a changelog entry? This might be confusing to some.

diniamo:
No clue why I called this autosave everywhere

NotAShelf (approved):

PR #521, opened 2025-01-05
statusline/feline: init

NotAShelf:
Adds a module for feline.nvim.

Mostly done, needs a module for compontents.active and compontents.inactive
before it is complete.


modules/plugins/statusline/feline/config.nix
```diff
@@ -0,0 +1,21 @@
...
+  ...
+}: let
+  inherit (lib.modules) mkIf;
+
+  cfg = config.vim.statusline.feline-nvim;
+in {
+  config = mkIf cfg.enable {
+    vim = {
+      lazy.plugins.feline-nvim = {
+        event = "UIEnter";
```
diniamo, line 13:
```suggestion
        event = "DeferredUIEnter";
```
This is debatable, but if the goal is to start accepting user input as soon as possible, then `DeferredUIEnter` should be used

modules/plugins/statusline/feline/feline.nix
```diff
@@ -0,0 +1,144 @@
...
+        description = "Buffer names in which to force render inactive statusline";
+      };
+    };
+  };
+in {
+  options.vim.statusline.feline-nvim = {
+    enable = mkEnableOption "minimal, stylish and customizable statusline, statuscolumn, and winbar [feline.nvim]";
+    setupOpts = mkPluginSetupOption "feline-nvim" {
+      custom_providers = mkOption {
+        type = attrsOf anything;
```
diniamo, line 48: Should be of type luaInline

modules/plugins/statusline/feline/feline.nix
```diff
@@ -0,0 +1,144 @@
...
+          Either a string containing the color theme name or an attribute set of
+          strings, containing the colors.
+
+          The theme’s `fg` and `bg` values also represent the default foreground
+          and background colors, respectively.
+        '';
+      };
+
+      separators = mkOption {
+        type = listOf str;
```
diniamo, line 80: Aaand wrong again, this should be an attribute set, ref https://github.com/freddiehaddad/feline.nvim/blob/9f1313f61a75ec5ebe805fedd46bdc130c420963/lua/feline/defaults.lua#L17-L38

modules/plugins/statusline/feline/feline.nix
```diff
@@ -0,0 +1,144 @@
...
+        type = attrsOf (submodule conditionalRenderers);
+        description = ''
+          A table that determines which buffers should always have the inactive
+          statusline, even when they are active.
+        '';
+      };
+
+      disable = mkOption {
+        default = {};
+        type = attrsOf (submodule conditionalRenderers);
```
diniamo, line 105: This doesn't make sense, the defaults above only apply to `force_active`. Nothing should be disabled by default.

modules/plugins/statusline/feline/feline.nix
```diff
@@ -0,0 +1,144 @@
...
+
+          It can later be used to get the color associated with the current Vim
+          mode using `require('feline.providers.vi_mode').get_mode_color()`.
+
+          See `:help feline-vi-mode` for more details on vi mode.
+        '';
+      };
+
+      # TODO:
+      # - Check for the number of elements in the list maybe? Can Feline have infinite components per section?
```
diniamo, line 126: Yes it can, there is no hard limit on it

modules/plugins/statusline/feline/feline.nix
```diff
@@ -0,0 +1,144 @@
...
+      # - type-check contents of `attrsOf anything`? Is there a downside for not doing that?
```
diniamo, line 127: I'm already against adding any options below setupOpts; that's just way overkill

diniamo (changes requested):

modules/plugins/statusline/feline/config.nix
NotAShelf: Is there a benefit to taking user input as soon as possible?

modules/plugins/statusline/feline/feline.nix
NotAShelf: `custom_providers` takes a table, in which case we should be able to get a `listOf luaInline`. No?

modules/plugins/statusline/feline/feline.nix
NotAShelf: We cannot omit the default here. That's why it has one, and also conditionalRenderers submodule provides upstream defaults.

modules/plugins/statusline/feline/config.nix
diniamo: I mean yeah? You wouldn't want a 10s startup time would you

modules/plugins/statusline/feline/feline.nix
diniamo: Men go read the docs, you have to be able to the custom providers somehow. It's a set

modules/plugins/statusline/feline/feline.nix
diniamo: It literally does not, nothing is disabled by default

PR #524, opened 2025-01-05
git/gitsigns: migrate to `setupOpts`

NotAShelf:
Adds `vim.git.gitsigns.setupOpts` for user-specified setup table. We do not provide any defaults, as we did not do so previously and users may add their own options easily.


Soliprem (approved):
tested and LGTM.

PR #545, opened 2025-01-10
notes/obsidian: remove default dir-value

LilleAila:
As the `obsidian.nvim` readme describes, the `dir` option is provided for backwards compatibility. The problem with this is that because `nvf` defines this value (*with* a default value), means that it overrides the `workspaces` option, making it impossible to use. Removing it would keep the same functionality, just without the default.

### DIFF
--- a/modules/plugins/notes/obsidian/obsidian.nix
+++ b/modules/plugins/notes/obsidian/obsidian.nix
@@ -24,12 +24,6 @@ in {
       enable = mkEnableOption "complementary neovim plugins for Obsidian editor";
 
       setupOpts = mkPluginSetupOption "Obsidian.nvim" {
-        dir = mkOption {
-          type = str;
-          default = "~/my-vault";
-          description = "Obsidian vault directory";
-        };
-
         daily_notes = {
           folder = mkOption {
             type = nullOr str;
--- a/docs/release-notes/rl-0.8.md
+++ b/docs/release-notes/rl-0.8.md
@@ -39,6 +39,5 @@
 
 [LilleAila](https://github.com/LilleAila):
 
-[obsidian.nvim](https://github.com/epwalsh/obsidian.nvim)
-
-- Remove `vim.notes.obsidian.setupOpts.dir`. Fixes issue with setting the workspace directory.
+- Remove `vim.notes.obsidian.setupOpts.dir`, which was set by default. 
+  Fixes issue with setting the workspace directory.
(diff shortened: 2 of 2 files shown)


modules/plugins/notes/obsidian/obsidian.nix
```diff
@@ -24,12 +24,6 @@ in {
       enable = mkEnableOption "complementary neovim plugins for Obsidian editor";
 
       setupOpts = mkPluginSetupOption "Obsidian.nvim" {
-        dir = mkOption {
-          type = str;
-          default = "~/my-vault";
-          description = "Obsidian vault directory";
-        };
```
NotAShelf, line 31: I believe you can change `default` to an `example` here and it'll still show up in the documentation without altering the setup table.

NotAShelf (commented):
LGTM. One small suggestion, but it's not strictly necessary.

Please do add an entry to the changelog before I can merge this.

modules/plugins/notes/obsidian/obsidian.nix
LilleAila: > I believe you can change `default` to an `example` here and it'll still show up in the documentation without altering the setup table.

That unfortunately wouldn't work, as it expects it to be defined, so it throws this error:
```
error: The option `vim.notes.obsidian.setupOpts.dir' was accessed but has no value defined. Try setting the option.
```

Another option would be this:
```nix
        dir = mkOption {
          type = nullOr str;
          example = "~/my-vault";
          default = null;
          description = "Obsidian vault directory";
        };
```
, but would that be a bit more messy? In my opinion, removing it looks better.

NotAShelf: null would be converted to `nil` in the resulting init.lua, and I'm not sure if obsidian.nvim actually handles that. Not to worry, you can simply remove the option.

LilleAila:
> Please do add an entry to the changelog before I can merge this.

Added to changelog, hopefully written correctly

modules/plugins/notes/obsidian/obsidian.nix
```diff
@@ -25,8 +25,9 @@ in {
 
       setupOpts = mkPluginSetupOption "Obsidian.nvim" {
         dir = mkOption {
-          type = str;
-          default = "~/my-vault";
+          type = nullOr str;
+          example = "~/my-vault";
+          default = null;
```
NotAShelf, line 30: Does `obsidian.nvim` handle this properly?

LilleAila: > Does `obsidian.nvim` handle this properly?

Yes, from my testing, although i accidentally commited it. Reverting it, as the other option seems better

NotAShelf:
Merging once CI passes.

NotAShelf:
thanks :)

PR #549, opened 2025-01-11
snippets/luasnip: add setupOpts

LilleAila:
Added setupOpts to `vim.snippets.luasnip`.

Right now, there is a soft dependency on nvim-cmp, because setting `lazy = true`, makes it so that it won't load if nvim-cmp is not present and configured. I will not fix that in this pr, as users can set `vim.lazy.plugins.luasnip.event = "BufEnter"` themselves.

### DIFF
--- a/docs/release-notes/rl-0.8.md
+++ b/docs/release-notes/rl-0.8.md
@@ -51,3 +51,5 @@
 
 - Remove `vim.notes.obsidian.setupOpts.dir`, which was set by default. Fixes
   issue with setting the workspace directory.
+- Add `vim.snippets.luasnip.setupOpts`, which was previously missing.
+- Add a trigger event for luasnip lazy-loading
--- a/docs/release-notes/rl-0.8.md
+++ b/docs/release-notes/rl-0.8.md
@@ -52,4 +52,3 @@
 - Remove `vim.notes.obsidian.setupOpts.dir`, which was set by default. Fixes
   issue with setting the workspace directory.
 - Add `vim.snippets.luasnip.setupOpts`, which was previously missing.
-- Add a trigger event for luasnip lazy-loading
--- a/modules/plugins/snippets/luasnip/config.nix
+++ b/modules/plugins/snippets/luasnip/config.nix
@@ -11,10 +11,13 @@ in {
     vim = {
       lazy.plugins.luasnip = {
         package = "luasnip";
+        
         lazy = true;
-        after = cfg.loaders;
+        
         setupModule = "luasnip";
         inherit (cfg) setupOpts;
+        
+        after = cfg.loaders;
       };
       startPlugins = cfg.providers;
       autocomplete.nvim-cmp = {
--- a/modules/plugins/snippets/luasnip/config.nix
+++ b/modules/plugins/snippets/luasnip/config.nix
@@ -9,12 +9,12 @@
 in {
   config = mkIf cfg.enable {
     vim = {
-      lazy.plugins = {
-        luasnip = {
-          package = "luasnip";
-          lazy = true;
-          after = cfg.loaders;
-        };
+      lazy.plugins.luasnip = {
+        package = "luasnip";
+        event = "BufEnter";
+        after = cfg.loaders;
+        setupModule = "luasnip";
+        inherit (cfg) setupOpts;
       };
       startPlugins = cfg.providers;
       autocomplete.nvim-cmp = {
--- a/modules/plugins/snippets/luasnip/config.nix
+++ b/modules/plugins/snippets/luasnip/config.nix
@@ -11,12 +11,12 @@ in {
     vim = {
       lazy.plugins.luasnip = {
         package = "luasnip";
-        
+
         lazy = true;
-        
+
         setupModule = "luasnip";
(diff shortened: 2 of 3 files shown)


NotAShelf (dismissed):

modules/plugins/snippets/luasnip/config.nix
```diff
@@ -9,12 +9,12 @@
...
     vim = {
-      lazy.plugins = {
-        luasnip = {
-          package = "luasnip";
-          lazy = true;
-          after = cfg.loaders;
-        };
+      lazy.plugins.luasnip = {
+        package = "luasnip";
+        event = "BufEnter";
```
diniamo, line 14:
```suggestion
        lazy = true;
```
The reason we use `lazy = true` is because LuaSnip is automatically loaded when cmp_luasnip requires it and I don't see a reason to change that since using LuaSnip through nvim-cmp is the only supported way built into nvf.

docs/release-notes/rl-0.8.md
```diff
@@ -51,3 +51,5 @@
 
 - Remove `vim.notes.obsidian.setupOpts.dir`, which was set by default. Fixes
   issue with setting the workspace directory.
+- Add `vim.snippets.luasnip.setupOpts`, which was previously missing.
+- Add a trigger event for luasnip lazy-loading
```
diniamo, line 55:
```suggestion
```
See below

diniamo (changes requested):

modules/plugins/snippets/luasnip/config.nix
LilleAila: > The reason we use `lazy = true` is because LuaSnip is automatically loaded when cmp_luasnip requires it, and I don't see a reason to change that since using LuaSnip through nvim-cmp is the only supported way built into nvf.

This causes a problem for people who do not use nvim-cmp though. It is not a hard dependency of luasnip, and it should be possible to use luasnip without completion, or with another plugin, such as blink, which i use.

diniamo: Blink support is being implemented internally, but if you don't use blink it is still possible to:
- use the cmp_luasnip provider with blink
- modify the LuaSnip lazy definition in your configuration (this is the reason it's an attribute set)

LilleAila: > Blink support is being implemented internally, but if you don't use cmp it is still possible to:
> 
>     * use the cmp_luasnip provider with blink
> 
>     * modify the LuaSnip lazy definition in your configuration (this is the reason it's an attribute set)

Seeing as LuaSnip doesn't provide the cmp source itself, but requires the cmp_luasnip plugin, i don't think they shouldn't be connected here by default either, especially considering blink has its own source for luasnip built in, which does not trigger the lazy loading.

> Expand LSP-Snippets with [nvim-compe](https://github.com/hrsh7th/nvim-compe) (or its' successor, [nvim-cmp](https://github.com/hrsh7th/nvim-cmp) (requires [cmp_luasnip](https://github.com/saadparwaiz1/cmp_luasnip)))

diniamo: Why not? How does it use luasnip then?

Also, another option is to trigger the load after you load blink with `require("lz.n").trigger_load("luasnip")`

horriblename: luasnip should just work with blink, remember to set `snippets.preset = 'luasnip'` in blink's config

LilleAila: > luasnip should just work with blink, remember to set `snippets.preset = 'luasnip'` in blink's config

That works as it should, it just doesn't trigger the lazy loading, i don't know the reason for that, but some other event (in my case BufEnter) had to trigger lazy loading for it to work

modules/plugins/snippets/luasnip/config.nix
```diff
@@ -9,12 +9,12 @@
...
-          package = "luasnip";
-          lazy = true;
-          after = cfg.loaders;
-        };
+      lazy.plugins.luasnip = {
+        package = "luasnip";
+        lazy = true;
+        after = cfg.loaders;
+        setupModule = "luasnip";
+        inherit (cfg) setupOpts;
```
diniamo, line 17:
```suggestion
        package = "luasnip";
        
        lazy = true;
        
        setupModule = "luasnip";
        inherit (cfg) setupOpts;
        
        after = cfg.loaders;
```
Sorry, but let's try to keep things consistent :b

diniamo (changes requested):
There else seem to be conflicts

diniamo (approved):

diniamo:
-.-

diniamo:
@NotAShelf preview action fail, seems to be unrelated

diniamo (approved):

NotAShelf:
I'm aware, refer to my comment on Discord for the CI. 

Can't review for the next 10 hours or so, feel free to merge if you feel it's ready.

diniamo:
Can't merge without your review apparently

NotAShelf (approved):

NotAShelf:
Thank you

PR #571, opened 2025-01-19
highlight: init

LilleAila:
Creates options for configuring highlight groups, mirroring the options available in `nvim_set_hl`.

### DIFF
--- a/docs/release-notes/rl-0.8.md
+++ b/docs/release-notes/rl-0.8.md
@@ -102,6 +102,7 @@
   - `mini.trailspace`
   - `mini.visits`
 - Add [fzf-lua](https://github.com/ibhagwan/fzf-lua) in `vim.fzf-lua`
+- Add options to define highlights under `vim.highlight`
 
 [kaktu5](https://github.com/kaktu5):
 
--- a/docs/release-notes/rl-0.8.md
+++ b/docs/release-notes/rl-0.8.md
@@ -102,7 +102,7 @@
   - `mini.trailspace`
   - `mini.visits`
 - Add [fzf-lua](https://github.com/ibhagwan/fzf-lua) in `vim.fzf-lua`
-- Add options to define highlights under `vim.highlight`
+- Add options to define highlights under [](#opt-vim.highlight)
 
 [kaktu5](https://github.com/kaktu5):
 
--- a/modules/neovim/init/highlight.nix
+++ b/modules/neovim/init/highlight.nix
@@ -3,11 +3,11 @@
   lib,
   ...
 }: let
-  inherit (lib.options) mkOption literalExpression;
+  inherit (lib.options) mkOption;
   inherit (lib.types) nullOr attrsOf listOf submodule bool ints str enum;
-  inherit (lib.strings) hasPrefix concatStringsSep;
+  inherit (lib.strings) hasPrefix concatLines;
   inherit (lib.attrsets) mapAttrsToList;
-  inherit (lib.nvim.dag) entryAnywhere;
+  inherit (lib.nvim.dag) entryBetween;
   inherit (lib.nvim.lua) toLuaObject;
   inherit (lib.nvim.types) hexColor;
 
@@ -15,18 +15,18 @@
     mkOption {
       type = nullOr hexColor;
       default = null;
+      example = "#ebdbb2";
       description = ''
         The ${target} color to use. Written as color name or hex "#RRGGBB".
       '';
-      example = "#ebdbb2";
     };
 
   mkBoolOption = name:
     mkOption {
       type = nullOr bool;
       default = null;
-      description = ''Whether to enable ${name}'';
       example = false;
+      description = "Whether to enable ${name}";
     };
 
   cfg = config.vim.highlight;
@@ -98,14 +98,14 @@ in {
       };
     });
     default = {};
-    description = "Custom highlight to apply";
-    example = literalExpression ''
+    example = ''
       {
         SignColumn = {
           bg = "#282828";
         };
       }
     '';
+    description = "Custom highlights to apply";
   };
 
   config = {
@@ -116,6 +116,6 @@ in {
         )
         cfg;
     in
-      entryAnywhere (concatStringsSep "\n" highlights);
+      entryBetween ["lazyConfigs" "pluginConfigs" "extraPluginConfigs"] ["theme"] (concatLines highlights);
   };
 }
--- a/modules/neovim/init/highlight.nix
+++ b/modules/neovim/init/highlight.nix
@@ -92,7 +92,7 @@ in {
             "NONE"
           ]));
           default = null;
-          description = "The cterm arguments to use. See :h highlight-args";
+          description = "The cterm arguments to use. See ':h highlight-args'";
         };
         force = mkBoolOption "force update";
       };
--- a/modules/neovim/init/highlight.nix
+++ b/modules/neovim/init/highlight.nix
@@ -4,7 +4,7 @@
   ...
 }: let
   inherit (lib.options) mkOption literalExpression;
-  inherit (lib.types) nullOr attrsOf listOf submodule bool ints str;
+  inherit (lib.types) nullOr attrsOf listOf submodule bool ints str enum;
   inherit (lib.strings) hasPrefix concatStringsSep;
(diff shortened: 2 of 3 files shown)


modules/neovim/init/highlight.nix
```diff
@@ -0,0 +1,121 @@
...
+    });
+    default = {};
+    description = "Custom highlight to apply";
+    example = literalExpression ''
+      {
+        SignColumn = {
+          bg = "#282828";
+        };
+      }
+    '';
```
NotAShelf, line 108: `literalExpression` is not necessary here, since there is nothing to be evaluated. It can be omitted.

modules/neovim/init/highlight.nix
```diff
@@ -0,0 +1,121 @@
...
+  inherit (lib.nvim.types) hexColor;
+
+  mkColorOption = target:
+    mkOption {
+      type = nullOr hexColor;
+      default = null;
+      description = ''
+        The ${target} color to use. Written as color name or hex "#RRGGBB".
+      '';
+      example = "#ebdbb2";
```
NotAShelf, line 21:
```suggestion
      default = null;
      example = "#ebdbb2";
      description = ''
        The ${target} color to use. Written as color name or hex "#RRGGBB".
      '';
```

modules/neovim/init/highlight.nix
```diff
@@ -0,0 +1,121 @@
...
+    };
+
+  mkBoolOption = name:
+    mkOption {
+      type = nullOr bool;
+      default = null;
+      description = ''Whether to enable ${name}'';
```
NotAShelf, line 28:
```suggestion
      description = "Whether to enable ${name}";
```

Also move `example` between `default` and `description` please.

modules/neovim/init/highlight.nix
```diff
@@ -0,0 +1,121 @@
...
+            "NONE"
+          ]));
+          default = null;
+          description = "The cterm arguments to use. See :h highlight-args";
+        };
+        force = mkBoolOption "force update";
+      };
+    });
+    default = {};
+    description = "Custom highlight to apply";
```
NotAShelf, line 101:
```suggestion
    description = "Custom highlights to apply";
```

I think highlight*s* would make more sense here, since we can set more than more highlight at once.

modules/neovim/init/highlight.nix
```diff
@@ -0,0 +1,121 @@
...
+
+  config = {
+    vim.luaConfigRC.highlight = let
+      highlights =
+        mapAttrsToList (
+          name: value: ''vim.api.nvim_set_hl(0, ${toLuaObject name}, ${toLuaObject value})''
+        )
+        cfg;
+    in
+      entryAnywhere (concatStringsSep "\n" highlights);
```
NotAShelf, line 119: `concatLines` can replace `concatStringsSep "\n"` here.

NotAShelf, line 119: `entryAnywhere` *should* be fine, but we might benefit more from placing this before plugin setup since some of them might actually use the highlights set in this module.

docs/release-notes/rl-0.8.md
```diff
@@ -102,6 +102,7 @@
   - `mini.trailspace`
   - `mini.visits`
 - Add [fzf-lua](https://github.com/ibhagwan/fzf-lua) in `vim.fzf-lua`
+- Add options to define highlights under `vim.highlight`
```
NotAShelf, line 105: You should be able to refer to the option with a documentation link with `[](#opt-vim.highlight)` since a product of `mkOption`

NotAShelf (changes requested):

modules/neovim/init/highlight.nix
```diff
@@ -0,0 +1,121 @@
...
+      '';
+    };
+
+  mkBoolOption = name:
+    mkOption {
+      type = nullOr bool;
+      default = null;
+      example = false;
+      description = "Whether to enable ${name}";
+    };
```
NotAShelf, line 30: Not a blocker, but these may be better suited for the extended lib.

modules/neovim/init/highlight.nix
```diff
@@ -0,0 +1,121 @@
...
+        standout = mkBoolOption "standout";
+        underline = mkBoolOption "underline";
+        undercurl = mkBoolOption "undercurl";
+        underdouble = mkBoolOption "underdouble";
+        underdotted = mkBoolOption "underdotted";
+        underdashed = mkBoolOption "underdashed";
+        strikethrough = mkBoolOption "strikethrough";
+        italic = mkBoolOption "italic";
+        reverse = mkBoolOption "reverse";
+        nocombine = mkBoolOption "nocombine";
```
NotAShelf, line 56: `mapAttrs (_: mkBoolOption)` can probably handle this with less repetition of code.

modules/neovim/init/highlight.nix
```diff
@@ -0,0 +1,121 @@
...
+            "reverse"
+            "inverse"
+            "italic"
+            "standout"
+            "altfont"
+            "nocombine"
+            "NONE"
+          ]));
+          default = null;
+          description = "The cterm arguments to use. See :h highlight-args";
```
NotAShelf, line 95:
```suggestion
          description = "The cterm arguments to use. See `:h highlight-args`";
```

modules/neovim/init/highlight.nix
```diff
@@ -0,0 +1,121 @@
...
+      };
+    });
+    default = {};
+    example = ''
+      {
+        SignColumn = {
+          bg = "#282828";
+        };
+      }
+    '';
```
NotAShelf, line 107:
```suggestion
    example = {
      SignColumn = {
        bg = "#282828";
      };
    };
```

Does not need to be a string literal.

modules/neovim/init/highlight.nix
```diff
@@ -0,0 +1,121 @@
...
+
+  config = {
+    vim.luaConfigRC.highlight = let
+      highlights =
+        mapAttrsToList (
+          name: value: ''vim.api.nvim_set_hl(0, ${toLuaObject name}, ${toLuaObject value})''
+        )
+        cfg;
+    in
+      entryBetween ["lazyConfigs" "pluginConfigs" "extraPluginConfigs"] ["theme"] (concatLines highlights);
```
NotAShelf, line 119: CC @horriblename on this. I'm not actually sure where `lazyConfigs` is located in the DAG system.

NotAShelf (changes requested):

modules/neovim/init/highlight.nix
LilleAila: I wrote it based on the code in `modules/wrapper/rc/config.nix`, where those three are listed as after each other.

modules/neovim/init/highlight.nix
LilleAila:
```nix
        // builtins.listToAttrs (map (option:
          nameValuePair option (mkOption {
            type = nullOr bool;
            default = null;
            example = false;
            description = "Whether to enable ${option}";
          })) [
          "bold"
          "standout"
          "underline"
          "undercurl"
          "underdotted"
          "underdashed"
          "strikethrough"
          "italic"
          "reverse"
          "nocombine"
          "force"
        ]);
```
Something like this? I don't really like that code, i think it looked better with repetition of `mkBoolOption`

NotAShelf:
```nix
mapAttrs (_: mkBoolOption) {
  bold = "bold";
  standout = "standout";
  underline = "underline";
  undercurl =  "undercurl";
  underdouble = "underdouble";
  underdotted = "underdotted";
  underdashed = "underdashed";
  strikethrough =  "strikethrough";
  italic = "italic";
  reverse =  "reverse";
  nocombine =  "nocombine";
}
```

would do it if I'm not mistaken. But you can keep the current method.

LilleAila: I feel that the code looks better with the repetition, and the two alternatives only introduce unneeded complexity and makes the code harder to read, even when there is slightly more repetition.

NotAShelf (dismissed):

NotAShelf:
If you can fix the conflicts, I can merge this.

NotAShelf (approved):

NotAShelf:
Thank you :)

PR #588, opened 2025-01-27
languages/typst: Add inline preview via typst-concealer

PartyWumpus:
Adds the [typst-concealer](https://github.com/PartyWumpus/typst-concealer) plugin for displaying typst inline.

This is marked as draft because I'm not sure about:
- [ ] Do I need to put something in some release notes?
- [ ] Should I add default keybinds? The plugin basically *needs* them.
- [ ] Should I change the descriptions to better fit the way they're phrased elsewhere? Currently they're just [ripped from my code](https://github.com/PartyWumpus/typst-concealer/blob/2f6a314736e0ee754c0014b5c63356ba6d9e9adc/lua/typst-concealer/init.lua#L696-L702).
- [ ] Should I be explicit about the defaults instead of just `null` and letting the plugin pick?


NotAShelf:
>  Do I need to put something in some release notes?

A very quick entry would be nice.

> Should I add default keybinds? The plugin basically needs them.

Sure, you can set them as its done in many module. See telescope for an example.

> Should I change the descriptions to better fit the way they're phrased elsewhere? Currently they're just ripped from my code.

Option descriptions are usually from plugin documentation in our case. You're free to change them, but I'm not going to ask you to.

> Should I be explicit about the defaults instead of just null and letting the plugin pick?

Leaving setupOpts empty means plugin's defaults will be used (if the plugin sets them.) Up to you whether you think some defaults are more... equal than others.

modules/plugins/languages/typst.nix
```diff
@@ -167,6 +167,48 @@ in {
...
         };
       };
+      typst-concealer = {
+        enable =
+          mkEnableOption ''
+            [typst-concealer]: https://github.com/PartyWumpus/typst-concealer
+
+            Inline typst preview for Neovim via [typst-concealer]
+          ''
+          // {default = false;};
```
NotAShelf, line 177:
```suggestion
          mkEnableOption ''
            [typst-concealer]: https://github.com/PartyWumpus/typst-concealer

            Inline typst preview for Neovim via [typst-concealer]
          '';
```

mkEnableOption defaults to false.

modules/plugins/languages/typst.nix
```diff
@@ -167,6 +167,48 @@ in {
...
+          styling_type = mkOption {
+            type = nullOr (enum ["simple" "none" "colorscheme"]);
+            description = ''What kind of styling should typst-concealer apply to your typst?'';
+            default = null;
+          };
+          ppi = mkOption {
+            type = nullOr int;
+            description = ''What PPI should typst render at. Plugin default is 300, typst's normal default is 144.'';
+            default = null;
+          };
```
NotAShelf, line 204: `null` default will be converted to its Lua equivalent, nil, if set. If you wish to set those options manually (not strictly necessary in the setupOpts format, it's freeform) then you should set appropriate values as `null` values are converted, not filtered.

Also I'd appreciate if you could older `mkOption`paramters as
1. type
2. default
3. (optional) example
4. description

For reference, description does not need `''...''` but can be set as `"..."` unless you are escaping double quotes inside the string literal. using `''''` adds unnecessary whitespace to resulting markdown.

NotAShelf (changes requested):

Soliprem (changes requested):
assuming you tested it and it works fine, this LGTM. Just need to fix conflicts with main @PartyWumpus

PartyWumpus:
Sorry it took so long to get back to this, I didn't notice the ping :sob:. 

Fixed up the PR, but I'm a little confused what to do about the "spelling error" here. Am I good to just add `typ` to the typos file or is there a more local way to opt out?
image

PartyWumpus:
oh and yes i've tested this locally with `typst.enable = true` & `typst.extensions.typst-concealer.enable = true`, and it works as it should.

NotAShelf:
> Fixed up the PR, but I'm a little confused what to do about the "spelling error" here. Am I good to just add typ to the typos file or is there a more local way to opt out?

Please add it to the typos config so that the CI doesn't get triggered each time we commit to the repo.

modules/plugins/languages/typst.nix
```diff
@@ -150,6 +151,57 @@ in {
...
+            type = str;
+            default = getExe pkgs.typst;
+            description = "Where should typst-concealer look for your typst binary?";
+            example = ''lib.getExe pkgs.typst'';
+          };
+          conceal_in_normal = mkOption {
+            type = nullOr bool;
+            default = null;
+            description = "Should typst-concealer still conceal when the normal mode cursor goes over a line.";
+          };
```
NotAShelf, line 202: `null` doesn't actually mean null in the traditional sense here. If you pass null to the default value, it will be converted to Lua's `nil` and that will be passed to your plugin options instead. If you want to use your plugin's defaults, simplfy don't define those options and the freeform module will allow using any value. Alternatively you can match the default values here IF you want to define the options for documentation purposes.

PartyWumpus: the plugin will use the default if `nil` is passed in. 

I still want to define the options so they show up in docs and stuff, I suppose I can make the defaults match the plugin manually instead of automatically if you'd prefer?

NotAShelf: No need. I assumed you were confused about the null conversion but seems it's intended. You may discard my comment, I'm going to merge now.

P.S. If you can squash your commits into one I can avoid squash merging :)

PartyWumpus: I don't trust myself not to fuck up the squash 💔, I'd rather you just squash merge if that's alright.

NotAShelf:
Thank you wumpus :)

PR #593, opened 2025-02-01
languages: refactor step 1

NotAShelf:
This is the first step to the dreaded language refactor. Adds necessary groundwork for refactoring the `languages.*` option modules latter on.

We would like to rely on conform-nvim and nvim-lint for modular and customizable formatting and linting, and then deprecate none-ls in the second step.

Goals:

- User configurable formatters and linters
- Decent interfaces to allow for internal usage

Things to do:

- [ ] Decide on an interface for configuring formatters
- [ ] Document new formatters and linters

Things done:

- [x] Add conform.nvim
- [ ] Add nvim-lint (WIP)


modules/plugins/formatter/conform-nvim/conform-nvim.nix
```diff
@@ -0,0 +1,56 @@
...
+          timeout_ms = 500;
+        };
+        description = ''
+          Table that will be passed to `conform.format()`. If this
+          is set, Conform will run the formatter on save.
+        '';
+      };
+
+      format_after_save = mkOption {
+        type = attrs;
```
horriblename, line 46: is format on save disable-able this way?

can we integrate our existing vim.g.formatOnSave? or should we remove it

NotAShelf: According to the [setup section](https://github.com/stevearc/conform.nvim?tab=readme-ov-file#setup) this just sets up an autocommand. Which we can *maybe* move out of setupOpts when I add an interface under `vim.autocommands`.

Implementation is as follows:

```lua
if opts.format_on_save then
    if type(opts.format_on_save) == "boolean" then
      opts.format_on_save = {}
    end
```

So we can simply can remove our implementation, or maybe use conform's [format API](https://github.com/stevearc/conform.nvim?tab=readme-ov-file#formatopts-callback) directly.

horriblename: make the type nullOr so we can disable it (tho it wouldn't matter once we get custom autocmd)

NotAShelf:
Closing, as changes from this PR already made it into main branch.

Issue #620, opened 2025-02-11
`npins` migration broke `blink-cmp`


alfarelcynthesis:
### ⚠️ Please verify that this bug has NOT been reported before.

- [x] I checked all existing issues and didn't find a similar issue

### Description

The blink configuration module references `blink-cmp` as a string, but it isn't fetched using npins, so trying to use it fails.

I don't quite understand the overlay structure of this project, so I wasn't able to fix it myself and pr a fix, but it should probably just be a matter of referencing the package directly instead of using the package name.

### 👟 Reproduction steps

Enable `vim.autocomplete.blink-cmp.enable`.

### 👀 Expected behavior

Build success and functioning plugin.

### 😓 Actual Behavior

Build failure.

### 💻 Metadata

 nix-shell -p nix-info --run "nix-info -m"

### 📝 Relevant log output

```bash
- system: `"x86_64-linux"`
 - host os: `Linux 6.12.9-xanmod1, NixOS, 25.05 (Warbler), 25.05.20250204.799ba5b`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Lix, like Nix) 2.91.1
System type: x86_64-linux
Additional system types: i686-linux, x86_64-v1-linux, x86_64-v2-linux, x86_64-v3-linux, x86_64-v4-linux
Features: gc, signed-caches
System configuration file: /etc/nix/nix.conf
User configuration files: /home/alfarel/.config/nix/nix.conf:/nix/store/bba927pia4zsz0phfzi68rg7dc2x6269-stylix-kde-config/nix/nix.conf:/etc/xdg/nix/nix.conf:/home/alfarel/.nix-profile/etc/xdg/nix/nix.conf:/nix/profile/etc/xdg/nix/nix.conf:/home/alfarel/.local/state/nix/profile/etc/xdg/nix/nix.conf:/etc/profiles/per-user/alfarel/etc/xdg/nix/nix.conf:/nix/var/nix/profiles/default/etc/xdg/nix/nix.conf:/run/current-system/sw/etc/xdg/nix/nix.conf
Store directory: /nix/store
State directory: /nix/var/nix
Data directory: /nix/store/gkdyw99c7fyzjsxd0f0haf84s77shm91-lix-2.91.1/share`
 - nixpkgs: `/nix/store/k8nkf470zpidpa5nh76lh2x6rxfzpwa4-source`
```

NotAShelf:
[...]

PR #632, opened 2025-02-12
flake: fix using blink-cmp package

Gerg-L:
Fixes: #620


alfarelcynthesis:
Thanks for fixing this, works flawlessly :).

PR #643, opened 2025-02-16
languages: inherit lspconfig options from language options

NotAShelf:
Converts all language modules to a newer, more robust `lsp.options` format that allows seamless overrides. This should be considered the first step of the grand language rework, where other options (diagnostics, formatters) will follow.

Adds `lib.lspOptions` to expose an extensible attribute set to provide the option table for each
LSP server template via lspconfig.

Converted languages:

- [x] Assembly (asm-lsp)
- [x] Astro
- [x] Bash
- [x] Clang
- [ ] Csharp
- [ ] CSS
- [x] Cue
- [x] Dart
- [ ] Gleam
- [ ] Go
- [ ] Haskell
- [ ] HCL
- [ ] Helm
- [ ] HTML
- [ ] Java
- [ ] Julia
- [ ] Kotlin
- [ ] Lua
- [ ] Markdown
- [ ] Nim
- [ ] Nix
- [ ] Nu
- [ ] OCaml
- [ ] Odin
- [ ] PHP
- [ ] Python
- [ ] R
- [ ] Ruby
- [ ] Rust
- [ ] Scala
- [ ] SQL
- [ ] Svelte
- [ ] Tailwind
- [ ] Terraform
- [ ] TS
- [ ] Typst
- [ ] Vala
- [ ] YAML
- [ ] Zig


lib/languages.nix
```diff
@@ -32,4 +34,31 @@ in {
       type = bool;
       description = "Turn on ${desc} for enabled languages by default";
     };
+
+  lspOptions = submodule {
```
horriblename, line 38: should probably just replace `vim.lsp.lspconfig.sources` with this, maybe

NotAShelf: As long as we're using lspconfig (which I don't think we'll be dropping until 0.11) lspconfig.sources should stay, in case users want to configure lspconfig manually.

NotAShelf: Yes that is correct.

lib/languages.nix
```diff
@@ -32,4 +34,31 @@ in {
...
+    options = {
+      capabilities = mkOption {
+        type = luaInline;
+        default = mkLuaInline "default_capabilities";
+        description = "LSP capabilitiess to pass to lspconfig";
+      };
+
+      on_attach = mkOption {
+        type = luaInline;
+        default = mkLuaInline "default_capabilities";
```
horriblename, line 49: typo?
```suggestion
        default = mkLuaInline "default_on_attach";
```

lib/languages.nix
```diff
@@ -32,4 +34,31 @@ in {
...
+        description = "Function to execute when an LSP server attaches to a buffer";
+      };
+
+      filetypes = mkOption {
+        type = listOf str;
+        description = "Filetypes to auto-attach LSP in";
+      };
+
+      cmd = mkOption {
```
horriblename, line 58: do you think it's necessary to split this into cmd, args? someone might want to override the cmd but keep the args

modules/plugins/languages/asm.nix
```diff
@@ -4,12 +4,34 @@
...
 
   cfg = config.vim.languages.assembly;
+
+  defaultServer = "asm-lsp";
+  servers = {
+    asm-lsp = {
+      package = pkgs.asm-lsp;
+      options = {
+        capabilities = mkLuaInline "capabilities";
+        on_attach = mkLuaInline "attach_keymaps";
```
horriblename, line 26: why not default_on_attach?

lib/languages.nix
```diff
@@ -32,4 +34,31 @@ in {
       type = bool;
       description = "Turn on ${desc} for enabled languages by default";
     };
+
+  lspOptions = submodule {
+    freeformType = attrsOf anything;
+    options = {
+      capabilities = mkOption {
+        type = luaInline;
+        default = mkLuaInline "default_capabilities";
```
horriblename, line 43: just "capabilities" I'm pretty sure

horriblename (changes requested):

lib/languages.nix
horriblename: nevermind, this creates more differnece between us and upstream config for little benefit

modules/plugins/languages/asm.nix
NotAShelf: Copied from a language module, and that was the default for the module I copied it from. Oversight on my behalf.

lib/languages.nix
NotAShelf: Right again...

modules/plugins/languages/bash.nix
```diff
@@ -126,7 +137,9 @@ in {
 
     (mkIf cfg.lsp.enable {
       vim.lsp.lspconfig.enable = true;
-      vim.lsp.lspconfig.sources.bash-lsp = servers.${cfg.lsp.server}.lspConfig;
+      vim.lsp.lspconfig.sources.bash-lsp = ''
+        lspconfig.("${cfg.lsp.server}").setup(${toLuaObject cfg.lsp.options})
```
horriblename, line 141: not valid lua? `lspconfig["${...}"].setup`?

modules/plugins/languages/clang.nix
```diff
@@ -5,48 +5,62 @@
...
-      '';
+      options = {
+        capabilities = mkLuaInline ''
+          {
+            offsetEncoding = { "utf-8", "utf-16" },
+            textDocument = {
+            completion = {
+              editsNearCursor = true
+            }
+          }
```
horriblename, line 53: this doesn't extend default capabilities does it? should do so otherwise snippets/other stuff won't work

modules/plugins/languages/bash.nix
NotAShelf: It should be valid lua

This'll evaluate to `lspconfig.("bash-ls").setup({ ... })`, which is valid. I've tested with Nix run, seems to be working but we might be able to `toLuaObject` the server too.

modules/plugins/languages/clang.nix
NotAShelf: This is the default value in lspconfig.

The situation is a bit annoying; since option values are converted *verbatim* using variables is out of the question, lest they turned into the following:

```lua
capabilities = local ccap = capabilities
               capabilities = ccap.extend
```

(Bad representation, but the point is that it has to be inline.)

NotAShelf:
```lua
capabilities = vim.tbl_deep_extend("force", capabilities, { offsetEncoding = { "utf-8", "utf-16" }, textDocument = { completion = { editsNearCursor = true } } })
```

Maybe like this? lmao

modules/plugins/languages/bash.nix
NotAShelf: That was in fact invalid. I was thinking of setup tables, fixed in HEAD

modules/plugins/languages/asm.nix
```diff
@@ -22,28 +44,46 @@ in {
...
-          on_attach = default_on_attach,
-          cmd = {"${cfg.lsp.package}/bin/asm-lsp"},
-        }
-      '';
+      vim.lsp.lspconfig = {
+        enable = true;
+        sources.asm-lsp = ''
+          lspconfig.${toLuaObject cfg.lsp.server}.setup(${toLuaObject cfg.lsp.options})
+        '';
+      };
```
LilleAila, line 86: Could something like this be handled directlly in the lspconfig module? Example from personal configuration: https://github.com/LilleAila/nvf-config/blob/main/config/languages/lspconfig.nix

, rather than having raw lua code in the modules themselves?

modules/plugins/languages/asm.nix
```diff
@@ -4,12 +4,34 @@
...
+    asm_lsp = {
+      package = pkgs.asm-lsp;
+      options = {
+        capabilities = mkLuaInline "capabilities";
+        on_attach = mkLuaInline "default_on_attach";
+        filetypes = ["asm" "vasm"];
+        cmd =
+          if isList cfg.lsp.package
+          then expToLua cfg.lsp.package
+          else ["${getExe cfg.lsp.package}"];
```
LilleAila, line 31: Would it be possible to make this be possibly `null`? When `cmd` is unset, it picks the LSP directly from PATH, which allows for installing them directly in devshells, but in this configuration, it is necessary to specify it.

modules/plugins/languages/asm.nix
NotAShelf: That is what I have in mind, instead of making LSP setups a part of language modules I intend to create a lower-level API to configure LSPs, possibly as `vim.lsp` or something similar. I would appreciate your feedback or/and proposals [here](https://github.com/NotAShelf/nvf/discussions/748)

modules/plugins/languages/asm.nix
NotAShelf: Yes, that's a good idea. The current idea was to set a `["lsp-cmd"]` to execute from PATH, but `null` might work if we filter those.

horriblename:
- with nvim-0.11 additions, we should provide default capabilities via `vim.lsp.config('*', {capabilities = ...})` 

- `default_on_attach` should maybe be moved to a lspconfig hook
  ```
      lspconfig.util.on_setup = lspconfig.util.add_hook_before(lspconfig.util.on_setup, function(config)
        if config.on_attach == nil then
          config.on_attach = default_on_attach
        end
      end)

  ```

note on on_attach: I haven't seen anything in nvim 0.11 that configures this aside from LspAttach autocmd, which may or may not be a problem if a language module needs to override a keybind, since idk if the order of autocmd is guranteed

with this, most language modules just need to set the LSP cmd and nothing else

sjcobb2022:
Is there anything i can do to help move this along?

I would love to have this configurability in my config

NotAShelf:
> Is there anything i can do to help move this along?
> 
> I would love to have this configurability in my config

I should've closed this PR, as it's superseded by https://github.com/NotAShelf/nvf/pull/801 in its entirety. The new `vim.lsp.servers` option is already in main branch if you're interested in configuring your own LSPs, but feel free to submit PRs to convert LSPs not covered by #801 if you would like to contribute

PR #651, opened 2025-02-18
git/git-conflict: avoid shadowing builtin keybinds

horriblename:
The current default git-conflict keybinds _conflicts_ with vim's builtin keybinds. This PR prefixes
them with the `<leader>` key


NotAShelf:
A changelog addition would be nice, this is a breaking change all things considered.

NotAShelf (approved):

PR #656, opened 2025-02-20
neovim/init: add autocommand & autogroup APIs

NotAShelf:
Adds `vim.augroups` and `vim.autocmds` as Nix-native APIs for defining autocommands and autogroups in Nix. The implementation is a bit hacky, and relies on Lua instead of Nix for resolving auto{cmd,group} tables but should work as intended. Basic testing yields no errors, but I'll know for sure after I convert all of my autocommands to the new format.

### DIFF
--- a/modules/neovim/init/autocmds.nix
+++ b/modules/neovim/init/autocmds.nix
@@ -1,0 +1,185 @@
+{
+  config,
+  lib,
+  ...
+}: let
+  inherit (lib.options) mkOption mkEnableOption literalExpression;
+  inherit (lib.lists) filter;
+  inherit (lib.strings) optionalString;
+  inherit (lib.types) nullOr submodule listOf str bool;
+  inherit (lib.nvim.types) luaInline;
+  inherit (lib.nvim.lua) toLuaObject;
+  inherit (lib.nvim.dag) entryAfter;
+
+  autocommandType = submodule {
+    options = {
+      enable =
+        mkEnableOption ""
+        // {
+          default = true;
+          description = "Whether to enable this autocommand";
+        };
+
+      event = mkOption {
+        type = nullOr (listOf str);
+        default = null;
+        example = ["BufRead" "BufWritePre"];
+        description = "The event(s) that trigger the autocommand.";
+      };
+
+      pattern = mkOption {
+        type = nullOr (listOf str);
+        default = null;
+        example = ["*.lua" "*.vim"];
+        description = "The file pattern(s) that determine when the autocommand applies).";
+      };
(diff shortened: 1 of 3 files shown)


horriblename (dismissed):

modules/neovim/init/autocmds.nix
```diff
@@ -0,0 +1,183 @@
...
+        type = nullOr (listOf str);
+        default = null;
+        example = ["*.lua" "*.vim"];
+        description = "The file pattern(s) that determine when the autocommand applies).";
+      };
+
+      callback = mkOption {
+        type = nullOr luaInline;
+        default = null;
+        example = ''
```
horriblename, line 40: perhaps add the `lib.generators.mkLuaInline` part?

horriblename: or make it a string and use `apply` to auto luaInline it

modules/neovim/init/autocmds.nix
```diff
@@ -0,0 +1,183 @@
...
+              }
+            )
+          end
+        '');
+      };
+    };
+
+    assertions = [
+      {
+        assertion = builtins.all (cmd: !(cmd.command != null && cmd.callback != null)) cfg.autocmds;
```
horriblename, line 178: is it required to have exactly one of them? if so, please use an xor

if not, I'd prefer `cmd.command == null || cmd.callback == null` but this is more a style nitpick

NotAShelf: Technically you can have an autocommand without either, but it doesn't make sense to.

modules/neovim/init/autocmds.nix
```diff
@@ -0,0 +1,183 @@
...
+        description = "The file pattern(s) that determine when the autocommand applies).";
+      };
+
+      callback = mkOption {
+        type = nullOr luaInline;
+        default = null;
+        example = ''
+          function()
+              print("Saving a Lua file...")
+          end,
```
horriblename, line 43:
```suggestion
          end
```

luaInline should be valid lua expressions

modules/neovim/init/autocmds.nix
horriblename: yea this is fine

modules/neovim/init/autocmds.nix
horriblename: nvim throws error if both are missing, xor this please

horriblename (changes requested):

horriblename (approved):

PR #672, opened 2025-02-28
plugins/utility/harpoon: added harpoon

UltraGhostie:
Harpoon (branch harpoon2) has been added as a plugin. Set under navigation to allow for more navigation plugins to be added in the future (set telescope under navigation?).
Basic options and keybinds are included.

First nvf and nix contribution, followed the guidelines as well as I could but tips and improvements appreciated :)

### DIFF
--- a/modules/plugins/utility/harpoon/config.nix
+++ b/modules/plugins/utility/harpoon/config.nix
@@ -1,0 +1,51 @@
+{
+  options,
+  config,
+  lib,
+  ...
+}: let
+  inherit (lib.modules) mkIf;
+  inherit (lib.lists) optionals;
+  inherit (lib.strings) concatStrings;
+  inherit (lib.nvim.binds) pushDownDefault mkKeymap;
+
+  cfg = config.vim.navigation.harpoon;
+
+  keys = cfg.mappings;
+  inherit (options.vim.navigation.harpoon) mappings;
+in {
+  config = mkIf cfg.enable {
+    vim = {
+      startPlugins = ["plenary-nvim"];
+
+      lazy.plugins.harpoon = {
+        package = "harpoon";
+        setupModule = "harpoon";
+        inherit (cfg) setupOpts;
+
+        before = ''
+        '';
+
+        after = ''
+          local harpoon = require("harpoon")
+          harpoon.setup({})
+        '';
+
+        cmd = ["Harpoon"];
+
+        keys = [
+          (mkKeymap "n" keys.markFile "<Cmd>lua require('harpoon'):list():add()<CR>" {desc = mappings.markFile.description;})
+          (mkKeymap "n" keys.listMarks "<Cmd>lua require('harpoon').ui:toggle_quick_menu(require('harpoon'):list())<CR>" {desc = mappings.listMarks.description;})
+          (mkKeymap "n" keys.file1 "<Cmd>lua require('harpoon'):list():select(1)<CR>" {desc = mappings.file1.description;})
+          (mkKeymap "n" keys.file2 "<Cmd>lua require('harpoon'):list():select(2)<CR>" {desc = mappings.file2.description;})
+          (mkKeymap "n" keys.file3 "<Cmd>lua require('harpoon'):list():select(3)<CR>" {desc = mappings.file3.description;})
+          (mkKeymap "n" keys.file4 "<Cmd>lua require('harpoon'):list():select(4)<CR>" {desc = mappings.file4.description;})
+        ];
+      };
+
+      binds.whichKey.register = pushDownDefault {
+        "<leader>a" = "Harpoon Mark";
+      };
+    };
+  };
+}
--- a/modules/plugins/utility/harpoon/default.nix
+++ b/modules/plugins/utility/harpoon/default.nix
@@ -1,0 +1,6 @@
+_: {
+  imports = [
+    ./harpoon.nix
+    ./config.nix
+  ];
+}
--- a/modules/plugins/utility/harpoon/harpoon.nix
+++ b/modules/plugins/utility/harpoon/harpoon.nix
@@ -1,0 +1,45 @@
+{lib, ...}: let
+  inherit (lib.options) mkEnableOption mkOption;
+  inherit (lib.types) bool;
+  inherit (lib.nvim.binds) mkMappingOption;
+  inherit (lib.nvim.types) mkPluginSetupOption luaInline;
+  inherit (lib.generators) mkLuaInline;
+  setupOptions = {
+    defaults = {
+      save_on_toggle = mkOption {
+        description = "any time the ui menu is closed then we will save the state back to the backing list, not to the fs";
+        type = bool;
+        default = false;
+      };
+      sync_on_ui_close = mkOption {
+        description = "any time the ui menu is closed then the state of the list will be sync'd back to the fs";
+        type = bool;
+        default = false;
+      };
+      key = mkOption {
+        description = "how the out list key is looked up. This can be useful when using worktrees and using git remote instead of file path";
+        type = luaInline;
+        default = mkLuaInline ''
+          function()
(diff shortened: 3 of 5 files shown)


modules/plugins/utility/harpoon/config.nix
```diff
@@ -0,0 +1,51 @@
...
+    vim = {
+      startPlugins = ["plenary-nvim"];
+
+      lazy.plugins.harpoon = {
+        package = "harpoon";
+        setupModule = "harpoon";
+        inherit (cfg) setupOpts;
+
+        before = ''
+        '';
```
NotAShelf, line 27: You do not need to set this if there is nothing to be done before the setup function.

modules/plugins/utility/harpoon/config.nix
```diff
@@ -0,0 +1,51 @@
...
+
+        after = ''
+          local harpoon = require("harpoon")
+          harpoon.setup({})
+        '';
```
NotAShelf, line 32: The `inherit (cfg) setupOpts;` line already handles this, you may safely omit this since the setup function will be constructed from `vim.navigation.harpoon` and added to the Lua configuration.

modules/plugins/utility/harpoon/default.nix
```diff
@@ -0,0 +1,6 @@
+_: {
```
NotAShelf, line 1:
```suggestion
{
```

No need for the empty argument placeholder, since this file will never receive arguments in the future.

modules/plugins/utility/harpoon/harpoon.nix
```diff
@@ -0,0 +1,45 @@
...
+      };
+      key = mkOption {
+        description = "how the out list key is looked up. This can be useful when using worktrees and using git remote instead of file path";
+        type = luaInline;
+        default = mkLuaInline ''
+          function()
+            return vim.loop.cwd()
+          end
+        '';
+      };
```
NotAShelf, line 27: Please order `mkOption` arguments as
1. type
2. default
3. (optional) example
4. description

This helps minimize diffs based on how likely an option is going to change. 

Additionally, it would better align with our style preferences if you could
- keep description lines to 80 characters per line. You can use multiple lines with multiline strings, e.g.:
```
description = ''
First line
Second line
''
```
Lastly, please capitalize the first letter of the description for `mkOption` description. `mkEnableOption` prepends "Whether to..." to the description, but this is not the case for `mkOption`.

modules/plugins/utility/harpoon/harpoon.nix
```diff
@@ -0,0 +1,45 @@
...
+    mappings = {
+      markFile = mkMappingOption "Mark file [Harpoon]" "<leader>a";
+      listMarks = mkMappingOption "List marked files [Harpoon]" "<C-e>";
+      file1 = mkMappingOption "Go to marked file 1 [Harpoon]" "<C-j>";
+      file2 = mkMappingOption "Go to marked file 2 [Harpoon]" "<C-k>";
+      file3 = mkMappingOption "Go to marked file 3 [Harpoon]" "<C-l>";
+      file4 = mkMappingOption "Go to marked file 4 [Harpoon]" "<C-;>";
+    };
+
+    enable = mkEnableOption "Harpoon: quick bookmarks on keybinds";
```
NotAShelf, line 41:
```suggestion
    enable = mkEnableOption "quick bookmarks on keybinds [Harpoon] ";
```

modules/plugins/utility/harpoon/harpoon.nix
```diff
@@ -0,0 +1,45 @@
...
+
+    setupOpts = mkPluginSetupOption "Harpoon" setupOptions;
```
NotAShelf, line 43: There is no need to store `setupOptions` in a variable here, since it is not likely to be reused. Putting the entire set into a `let in` block introduces a small overhead in performance, although negligible. You can define it here as `setupOpts = mkPluginSetupOption "Harpoon" { ... }` directly.

NotAShelf (changes requested):
Looks good to me overall! I've requested some changes to remain consistent with our stylistic choices, and to remain consistent with wording, code styles etc. If you could also add a changelog entry, this'll be good to merge.

modules/plugins/utility/harpoon/harpoon.nix
UltraGhostie: Descriptions are slightly weirdly formatted in file after `nix fmt` but I suppose that is expected

modules/plugins/utility/harpoon/harpoon.nix
```diff
@@ -0,0 +1,47 @@
...
+        };
+        key = mkOption {
+          type = luaInline;
+          default = mkLuaInline ''
+            function()
+              return vim.loop.cwd()
+            end
+          '';
+          description = ''            How the out list key is looked up. This can be useful
+                    when using worktrees and using git remote instead of file path'';
```
NotAShelf, line 42: The descriptions indicate alejandra did something wrong with formatting. Please try to format them as

```
''
text
''
```

NotAShelf:
Left one quick comment, LGTM otherwise.

modules/plugins/utility/harpoon/harpoon.nix
```diff
@@ -0,0 +1,53 @@
...
+    enable = mkEnableOption "Quick bookmarks on keybinds [Harpoon]";
+
+    setupOpts = mkPluginSetupOption "Harpoon" {
+      defaults = {
+        save_on_toggle = mkOption {
+          type = bool;
+          default = false;
+          description = ''
+            Any time the ui menu is closed then we will save the
+                      state back to the backing list, not to the fs
```
NotAShelf, line 27:
```suggestion
            Any time the ui menu is closed then we will save the
            state back to the backing list, not to the fs
```

modules/plugins/utility/harpoon/harpoon.nix
```diff
@@ -0,0 +1,53 @@
...
+          '';
+        };
+        sync_on_ui_close = mkOption {
+          type = bool;
+          default = false;
+          description = ''
+            Any time the ui menu is closed then the state of the
+                      list will be sync'd back to the fs
```
NotAShelf, line 35:
```suggestion
            Any time the ui menu is closed then the state of the
            list will be sync'd back to the fs
```

modules/plugins/utility/harpoon/harpoon.nix
```diff
@@ -0,0 +1,53 @@
...
+        key = mkOption {
+          type = luaInline;
+          default = mkLuaInline ''
+            function()
+              return vim.loop.cwd()
+            end
+          '';
+          description = ''
+            How the out list key is looked up. This can be useful
+                      when using worktrees and using git remote instead of file path
```
NotAShelf, line 47:
```suggestion
            How the out list key is looked up. This can be useful
            when using worktrees and using git remote instead of file path
```

NotAShelf (changes requested):
Sorry for all the nitpicks, last change I promise :)

UltraGhostie:
> Sorry for all the nitpicks, last change I promise :)

Don't worry, I understand the benefits of uniform code style :)

NotAShelf (dismissed):

NotAShelf:
The editorconfig check seems to be failing.

https://github.com/NotAShelf/nvf/actions/runs/13677849178/job/38253940120?pr=672

Please save the changelog with editorconfig enabled.

UltraGhostie:
Installing EditorConfig for VSCode with workspace being in the top directory and re-saving `rl-0.8.md` does not seem to do anything, no change that Git can detect at least.

NotAShelf (approved):

NotAShelf:
There is a small formatting woe in the diff, but at this point I could care less. I'm more interested in getting Harpoon in the module system than a nitpick.

Thank you for your efforts!

Issue #675, opened 2025-03-01
Delayed L key


butzist:
### ⚠️ Please verify that this bug has NOT been reported before.

- [x] I checked all existing issues and didn't find a similar issue

### Description

When navigating in normal mode, the execution of the L key (right, by default) is delayed by about a second. I was able to trace this down to setting `programs.nvf.settings.vim.lsp.nvim-docs-view.enable = true`.

### 👟 Reproduction steps

- Set `programs.nvf.settings.vim.lsp.nvim-docs-view.enable = true`
- Press the L key (lowercase) in normal mode

### 👀 Expected behavior

Move caret to right immediately

### 😓 Actual Behavior

Movement is delayed by about 1s

### 💻 Metadata

 - system: `"x86_64-linux"`  - host os: `Linux 6.13.4, NixOS, 25.05 (Warbler), 25.05.20250224.0196c01`  - multi-user?: `yes`  - sandbox: `yes`  - version: `nix-env (Nix) 2.24.12`  - nixpkgs: `/nix/store/593xvgv994xlkm5mb7w4p1xxnzrs9wv6-source`

### 📝 Relevant log output

```bash
nothing relevant
```

NotAShelf:
Thank you for the report.

Looking into the code, I think it's because hijack `lvt` and `lvu` as keybinds when it should be `<leader>lvt` and `<leader>lvu`. I'm currently away from my main system and cannot push a fix, but I would welcome a PR for this.

Alternatively, you can change the keybindins in your configuration manually.

butzist:
Agree, will create a PR

NotAShelf:
Closed by #677

PR #677, opened 2025-03-01
plugins/nvim-docs-view: fix default mappings (#675)

butzist:
Default mappings of plugin nvim-docs-view conflict with directional navigation in normal mode. This PR prefixes the keybindings with the leader key.


NotAShelf (approved):
LGTM, thank you!

PR #676, opened 2025-03-01
languages/yaml: Init

TheMaxMur:
## Sanity Checking

[editorconfig]: https://editorconfig.org
[changelog]: https://github.com/NotAShelf/nvf/tree/main/docs/release-notes
[hacking nvf]: https://notashelf.github.io/nvf/index.xhtml#sec-guidelines

- [x] I have updated the [changelog] as per my changes
- [x] I have tested, and self-reviewed my code
- [ ] My changes fit guidelines found in [hacking nvf]
- Style and consistency
  - [x] I ran **Alejandra** to format my code (`nix fmt`)
  - [x] My code conforms to the [editorconfig] configuration of the project
  - [x] My changes are consistent with the rest of the codebase
- If new changes are particularly complex:
  - [ ] My code includes comments in particularly complex areas
  - [ ] I have added a section in the manual
  - [ ] _(For breaking changes)_ I have included a migration guide
- Package(s) built:
  - [x] `.#nix` _(default package)_
  - [x] `.#maximal`
  - [x] `.#docs-html` _(manual, must build)_
  - [x] `.#docs-linkcheck` _(optional, please build if adding links)_
- Tested on platform(s)
  - [x] `x86_64-linux`
  - [ ] `aarch64-linux`
  - [ ] `x86_64-darwin`
  - [ ] `aarch64-darwin`

---

Add a :+1: [reaction] to [pull requests you find important].

[reaction]: https://github.blog/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/
[pull requests you find important]: https://github.com/NixOS/nixpkgs/pulls?q=is%3Aopen+sort%3Areactions-%2B1-desc

PR #679, opened 2025-03-01
languages/helm: Initial Helm chart language support

butzist:
Depends on #676 and #680

Adds language support for Helm charts via `helm-ls` and `vim-helm`

### DIFF
--- a/docs/release-notes/rl-0.8.md
+++ b/docs/release-notes/rl-0.8.md
@@ -243,3 +243,7 @@
 [BANanaD3V](https://github.com/BANanaD3V):
 
 - `alpha` is now configured with nix.
+
+[Butzist](https://github.com/butzist)
+
+- Add Helm chart support under `vim.languages.helm`.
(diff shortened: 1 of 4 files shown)


butzist:
Will rebase once #676  is merged

NotAShelf:
#676 is now merged.

butzist:
Rebased, but should be probably merged after #680

butzist:
@NotAShelf Should be ready for review now

butzist:
If I understand the issue correctly, the filetype is first set to yaml, which starts the yamlls, and then later adjusted to ft=helm, which starts the second lsp. The `ft` arg to lazy is supposed to only load the plugin when a specific filetype is detected. Not sure how this solves the issue, though.
 
 If we can't reproduce a similar behaviour, then probably killing the wrong lsp is the best solution. At least I tested it in both configurations: yaml only and yaml+helm

horriblename:
just a hunch but I'm guessing this is caused by mnw appending plugins to rtp. will dig around abit later

butzist:
@NotAShelf Can we then keep the solution as-is, and refine if we find a better one? For me it works fine so far.
Any other changes that I should address?

butzist:
Ready, thank you for the review

PR #687, opened 2025-03-05
Hide conform and nvim-lint

horriblename:
fixes nvim-lint options and hides conform and nvim-lint, since they're both not ready for use


LilleAila (commented):
Is there really any reason to add `internal = true` to the modules? They work perfectly fine for me, so i don't see any reason of doing that, even though `null-ls` is still the default. Otherwise LGTM.

NotAShelf:
bc978c4fad0645a2cc010783b2f411b8800e9a37 fixes the invalid setup table, so I'm closing this. We will need to re-visit nvim-lint regardless, but I do not have the energy for it right now. PRs welcome, but the error should no longer occur so I'm closing in the hopes of making the linter packages *declarative* in the future.

PR #691, opened 2025-03-07
Add a working copilot plugin.

anicolao:
The `copilot-lua` plugin that is included with `nvf` doesn't work at all for me, while the official plugin works excellently. Add a `"copilot"` option so that users can use the official plugin if they like.

Also fix a typo reported by me as a bug in the `npins` JSON because without that fix the `npins` doesn't work.

### DIFF
--- a/npins/sources.json
+++ b/npins/sources.json
@@ -279,6 +279,21 @@
       "url": "https://github.com/stevearc/conform.nvim/archive/a6f5bdb78caa305496357d17e962bbc4c0b392e2.tar.gz",
       "hash": "1jkm8pbfnp2s9y70cc67pj2fa25a4jl1y4lx6y1k5i323f4lplhz"
     },
+    "copilot": {
+      "type": "GitRelease",
+      "repository": {
+        "type": "GitHub",
+        "owner": "github",
+        "repo": "copilot.vim"
+      },
+      "pre_releases": false,
+      "version_upper_bound": null,
+      "release_prefix": null,
+      "version": "v1.43.0",
+      "revision": "5015939f131627a6a332c9e3ecad9a7cb4c2e549",
+      "url": "https://api.github.com/repos/github/copilot.vim/tarball/v1.43.0",
+      "hash": "198mymn0ad1kzqzs1rimrhq7z3nrami45fp3yqnwp15am8bxmwi0"
+    },
     "copilot-cmp": {
       "type": "Git",
       "repository": {


NotAShelf:
There is no need to modify the pins here if we are not adding a module. nvf provides an extra plugins API that you can use to install and configure plugins without relying on upstream (us.) I would encourage using said API, or refactoring to also include a module.

Note that I am not too keen on vimscript plugins. copilot.vim is very slow, and is not designed with modern setups in mind. Copilot.lua is supposed to be feature complete, but if I recall correctly the author was not able to work on bugfixes. Perhaps there is a fork that we can use?

P.S. the npins bug should be fixed with 414c92276efd552e0491eca8e20c4a68acfec255

anicolao:
> There is no need to modify the pins here if we are not adding a module. nvf provides an extra plugins API that you can use to install and configure plugins without relying on upstream (us.) I would encourage using said API, or refactoring to also include a module.
> 
> Note that I am not too keen on vimscript plugins. copilot.vim is very slow, and is not designed with modern setups in mind. Copilot.lua is supposed to be feature complete, but if I recall correctly the author was not able to work on bugfixes. Perhaps there is a fork that we can use?

I couldn't figure out how to use `extraPlugins` from the documentation. The example reads


```
    config.vim.extraPlugins = with pkgs.vimPlugins; {                             
      aerial = {                                                                  
        package = aerial-nvim;                                                    
        setup = "require('aerial').setup {}";                                     
      };                                                                          
                                                                                  
      harpoon = {                                                                 
        package = harpoon;                                                        
        setup = "require('harpoon').setup {}";                                    
        after = ["aerial"]; # place harpoon configuration after aerial            
      };                                                                          
    }     
```

and I failed to find any information on how to define a `copilot-nvim` at the user level that could be used as the `package` parameter and point at github the way the pin does (or some other way). I did spend a few hours on that approach before giving up and using the `npins` approach which even with the bug that I stumbled upon while doing it took me only about 45m to figure out and test.

There's an additional problem with this PR that I realized when I woke up this morning though which is that it depends on `nodejs` being in the user's environment. I don't know how/where to indicate that if the user uses this plugin `nodejs` should be included in the resulting neovim dependencies so that it comes from there rather than from the user's system. 

Maybe we can fix the documentation instead so that I could understand how to use `extraPlugins` directly; I'm a complete nix n00b and find the existing documentation pretty much impenetrable in spite of all the great effort that has been put in to make it easy/well documented.

NotAShelf:
In this example:

```nix
config.vim.extraPlugins = with pkgs.vimPlugins; {                             
      aerial = {                                                                  
        package = aerial-nvim;                                                    
        setup = "require('aerial').setup {}";                                     
      };                                                                          
                                                                                  
      harpoon = {                                                                 
        package = harpoon;                                                        
        setup = "require('harpoon').setup {}";                                    
        after = ["aerial"]; # place harpoon configuration after aerial            
      };                                                                          
    }     

```


the packages are acquired from the `with` scope. E.g. `package = harpoon` is `package = pkgs.vimPlugins.harpoon`. In hindsight this is a very confusing example, because `with scopes` are confusing by default. Perhaps I'll document using npins in your own configuration, but first it is worth clearing where the packages come from.

Looking at it now, neither copilot plugin seems to be packaged in Nixpkgs. You could request that they're packaged, but it is also possible that we add a module instead. Then again, I am not a huge fan of LLM companions and especially not a fan of Vimscript plugins.

anicolao:
https://github.com/NotAShelf/nvf/discussions/609

looks relevant. It isn't the `with` that was confusing, it was where/how to fetch the plugin from github, which is from my PoV what `npins` is doing for me, setting up a configuration that enables `nvim` to find the plugin. I can find no similar example in the documentation for how to do it. Possibly I can get it to go after reading that discussion more thoroughly.

I shoudl perhaps be 100% clear that I tried initially to use the "assistant" module to enable copilot, but aside from the fact that it dragged in a super irritating `copilot-cmp` configuration that I don't want, it also didn't work at all. Probably I configured it wrong. I could file a separate issue for that.

PR #699, opened 2025-03-09
Initial alpha.nvim nix implementation

BANanaD3V:
Allows to configure alpha with nix. Supports predefined themes aswell as setting a custom layout and opts.

TODO:
- Example config
- Typed layout

### DIFF
--- a/modules/plugins/dashboard/alpha/config.nix
+++ b/modules/plugins/dashboard/alpha/config.nix
@@ -5,8 +5,11 @@
 }: let
   inherit (lib.modules) mkIf;
   inherit (lib.nvim.dag) entryAnywhere;
+  inherit (lib.nvim.lua) toLuaObject;
 
   cfg = config.vim.dashboard.alpha;
+  themeDefined = cfg.theme != null;
+  layoutDefined = cfg.layout != [];
 in {
   config = mkIf cfg.enable {
     vim.startPlugins = [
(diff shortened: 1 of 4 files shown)


modules/plugins/dashboard/alpha/config.nix
```diff
@@ -5,216 +5,43 @@
...
-      		dashboard.button("SPC ff", "󰊄  Live grep"), -- telescope
-      		dashboard.button("SPC p", "  Projects"), -- any project
-      		dashboard.button("q", "󰅚  Quit", "<cmd>qa<CR>"), -- available all the time
-      	},
-      	position = "center",
+    assertions = [
+      {
+        assertion = themeDefined || layoutDefined;
+        message = ''
+          You should either define `theme` or `layout`.
```
NotAShelf, line 36:
```suggestion
          One of 'theme' or 'layout' should be defined in Alpha configuration.
```

Assertion messages don't contain formatting options, so using backticks is unnecessary.

modules/plugins/dashboard/alpha/config.nix
```diff
@@ -5,216 +5,43 @@
...
-      		setup = function()
-      			vim.cmd([[
-                  autocmd alpha_temp DirChanged * lua require('alpha').redraw()
-                  ]])
-      		end,
-      	},
+      {
+        assertion = !(themeDefined && layoutDefined);
+        message = ''
+          You can't define both `theme` and `layout` at the same time.
```
NotAShelf, line 42:
```suggestion
         'theme' and 'layout' cannot be defined at the same time.
```

NotAShelf (changes requested):
A few stylistic nitpicks for things that caught my attention.

All things considered this is a breaking change, so you must add an entry to the changelog under the breaking changes section.

Additionally, I think it should be made possible for users to use the old configuration somehow, but I can't think of a way that doesn't duplicate code unnecessarily so let's just mention the breakage and move on.

BANanaD3V:
> All things considered this is a breaking change, so you must add an entry to the changelog under the breaking changes section.

Sure. Will do.

> Additionally, I think it should be made possible for users to use the old configuration somehow, but I can't think of a way that doesn't duplicate code unnecessarily so let's just mention the breakage and move on.

I can try to recreate the old config as a default option for layout, but I think theme is already enough

NotAShelf (changes requested):
Please don't.

BANanaD3V:
oh no

NotAShelf (approved):

NotAShelf:
merging with a typo, wild

PR #743, opened 2025-03-24
fix: Update npins version

Gerg-L:
:rocket:


NotAShelf (approved):
who let the frog cook

NotAShelf:
if any of the workflows fail I'm going to defenestrate you

PR #755, opened 2025-03-29
theme: Add tinted-vim plugin theming support

JamyGolden:
[Tinted Theming](https://github.com/tinted-theming) is the community successor to Base16 and now supports multiple scheme systems, including Base24. This adds [tinted-vim](https://github.com/tinted-theming/tinted-vim) which is simply a bunch of vim colorschemes which can switches to through `vim.cmd.colorscheme("base16-ayu-dark")` for example.

This is my first PR so I'm sure I've done things wrong, but I have read `CONTRIBUTING.md`. A few things of note here:

- I ran `nix run .#maximal -Lv` but it didn't load with the plugin so I wasn't able to test my changes. How can I test the plugin which functions similarly to `base16-nvim`?
- I manually added the values to `npin.sources.json`, is this correct?

### DIFF
--- a/docs/release-notes/release-notes.md
+++ b/docs/release-notes/release-notes.md
@@ -3,6 +3,10 @@
 This section lists the release notes for tagged version of **nvf** and the
 current main current main branch
 
+[JamyGolden](https://github.com/JamyGolden):
+
+- Add support for `tinted-vim` (base16, base24) themes
+
 ```{=include=} chapters
 rl-0.1.md
 rl-0.2.md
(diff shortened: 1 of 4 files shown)


NotAShelf:
As far as I remember the current plugin doesn't actually support Base24 themes (which is a good improvement over Base16) but I am not a huge fan of adding more Vimscript plugins. In addition, the current implementation looks a little fragile.

```
setup = {style ? "base24-ayu-dark", ...}: ''
```

What if the user sets a style that is not actually available? `?` (basically hasAttr) is not type-safe either. I'm a little conflicted on adding this. Could there be a better plugin or/and a solution?

JamyGolden:
Thanks for the responses. I’ll look into creating a tinted-nvim repo built with lua and reopen this when things are looking good. 

As @NotAShelf mentioned it’s not just base16, but includes base24 and will contain themes from any other scheme system that is added in future.

PR #758, opened 2025-03-29
utility/oil-nvim: init

### LINKED ISSUE #722: Add: oil.nvim

- [x] I checked and didn't find a similar feature request

### 🏷️ Feature Type

New Addon

### 🔖 Feature description

Hi, it'd be nice if [oil.nvim](https://github.com/stevearc/oil.nvim) was available on the config, probably under the `vim.filetree` section

### ✔️ Solution

I would love to write this myself but im very new to nix, might still try to do so if i have enough time though!

### ❓ Alternatives

_No response_

### 📝 Additional Context

_No response_

NotAShelf:
Add oil.nvim, the alternative file explorer to edit filesystems as a buffer.

Closes #722

### DIFF
--- a/docs/release-notes/rl-0.8.md
+++ b/docs/release-notes/rl-0.8.md
@@ -26,6 +26,7 @@
 [yanky.nvim]: https://github.com/gbprod/yanky.nvim
 [yazi.nvim]: https://github.com/mikavilpas/yazi.nvim
 [snacks.nvim]: https://github.com/folke/snacks.nvim
+[oil.nvim]: https://github.com/stevearc/oil.nvim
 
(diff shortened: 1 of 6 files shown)


Soliprem:
would consider adding a default bind for `-`  to `:Oil<CR>` in normal mode

NotAShelf:
I'm not sure if we should keep adding more default binds. It has been a constant problem, and more default binds is more to refactor in the future when we get to it.

Soliprem:
what about an example? Only proposing it because it's what they do in the oil docs as well

NotAShelf:
I think that's for the user to decide, we shouldn't be the middleman between a plugin and its documentation.

PR #775, opened 2025-04-02
motion/flash-nvim: init

ckoehler:
A motion plugin, from Folke. https://github.com/folke/flash.nvim

Tested it locally and seems to work alright :) 

I'd like to be able to define additional setup opts, ideally as just a set to be converted into Lua; but I couldn't get it to work. Any help would be appreciated (could be a follow-on PR).

### DIFF
--- a/docs/release-notes/rl-0.8.md
+++ b/docs/release-notes/rl-0.8.md
@@ -301,6 +301,9 @@
 - Update default `telescope.setupOpts.pickers.find_files.find_command` to only
   include files (and therefore exclude directories from results)
 
-  [ckoehler](https://github.com/ckoehler):
+[ckoehler](https://github.com/ckoehler):
 
-  - Fix oil config referencing snacks
+[flash.nvim]: https://github.com/folke/flash.nvim
+
+- Fix oil config referencing snacks
+- Add [flash.nvim] plugin to `vim.utility.motion.flash-nvim`
--- a/modules/plugins/utility/motion/flash/config.nix
+++ b/modules/plugins/utility/motion/flash/config.nix
@@ -1,0 +1,34 @@
+{
+  config,
+  lib,
+  ...
+}: let
+  inherit (lib.modules) mkIf;
+  inherit (lib.nvim.binds) mkKeymap;
+  inherit (lib.nvim.lua) toLuaObject;
+  inherit (lib.nvim.dag) entryAnywhere;
+
+  cfg = config.vim.utility.motion.flash-nvim;
+in {
+  config = mkIf cfg.enable {
+    vim = {
+      lazy.plugins = {
+        "flash-nvim" = {
+          package = "flash-nvim";
+          setupModule = "flash";
+          setupOpts = cfg.setupOpts;
+
+          lazy = true;
+
+          keys = [
+            (mkKeymap ["n" "o" "x"] cfg.mappings.jump "<cmd>lua require(\"flash\").jump()<cr>" {desc = "Flash";})
+            (mkKeymap ["n" "o" "x"] cfg.mappings.treesitter "<cmd>lua require(\"flash\").treesitter()<cr>" {desc = "Flash Treesitter";})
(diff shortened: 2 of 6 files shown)


modules/plugins/utility/motion/flash/default.nix
```diff
@@ -0,0 +1,6 @@
+_: {
```
NotAShelf, line 1:
```suggestion
{
```

No need for an empty arg in files unlikely to get args.

modules/plugins/utility/motion/flash/flash.nix
```diff
@@ -0,0 +1,36 @@
...
+        type = nullOr str;
+        description = "Treesitter Search";
+        default = "R";
+      };
+      toggle = mkOption {
+        type = nullOr str;
+        description = "Toggle Flash Search";
+        default = "<c-s>";
+      };
+    };
```
NotAShelf, line 34: Please order `mkOption` arguments as 

1. type
2. default
3. (optional) example
4. deescription

modules/plugins/utility/motion/flash/flash.nix
```diff
@@ -0,0 +1,36 @@
+{lib, ...}: let
+  inherit (lib.options) mkEnableOption mkOption;
+  inherit (lib.types) nullOr str;
+in {
+  options.vim.utility.motion.flash = {
+    enable = mkEnableOption "flash.nvim plugin";
```
NotAShelf, line 6: Would be nice to describe what the plugin actually does here. For most plugin the format is

```nix
mkEnableOption "something something [flash.nvim]";

``` or similar while accounting for the fact that `mkEnableOption` prepends "Whether to enable" to the description.

File: docs/release-notes/rl-0.8.md [hunk 4]
```diff
@@ -301,6 +301,7 @@
 - Update default `telescope.setupOpts.pickers.find_files.find_command` to only
   include files (and therefore exclude directories from results)
 
-  [ckoehler](https://github.com/ckoehler):
+[ckoehler](https://github.com/ckoehler):
 
-  - Fix oil config referencing snacks
+- Fix oil config referencing snacks
+- Add `flash` plugin to `vim.utility.motion.flash`
```
NotAShelf, line 307: Could you include a link to the plugin page here? There are examples for non-inline markdown references above in this document.

NotAShelf (changes requested):
Some stylistic changes, LGTM otherwise.

NotAShelf:
Almost forgot, please also rename the module to `flash-nvim` instead of `flash`.

ckoehler:
> Almost forgot, please also rename the module to `flash-nvim` instead of `flash`.

Done. This also allowed me to use the additional setupOpts, which I added and tested. Thanks!

modules/plugins/utility/motion/flash/config.nix
```diff
@@ -0,0 +1,37 @@
...
+          keys = [
+            (mkKeymap ["n" "o" "x"] cfg.mappings.jump "<cmd>lua require(\"flash\").jump()<cr>" {desc = "Flash";})
+            (mkKeymap ["n" "o" "x"] cfg.mappings.treesitter "<cmd>lua require(\"flash\").treesitter()<cr>" {desc = "Flash Treesitter";})
+            (mkKeymap "o" cfg.mappings.remote "<cmd>lua require(\"flash\").remote()<cr>" {desc = "Remote Flash";})
+            (mkKeymap ["o" "x"] cfg.mappings.treesitter_search "<cmd>lua require(\"flash\").treesitter_search()<cr>" {desc = "Treesitter Search";})
+            (mkKeymap "c" cfg.mappings.toggle "<cmd>lua require(\"flash\").toggle()<cr>" {desc = "Toggle Flash Search";})
+          ];
+        };
+      };
+    };
```
NotAShelf, line 35: Any reason we're not passing cfg.setupOpts to lazy.plugins."flash-nvim".setupOpts? That should handle the setup call automatically.

NotAShelf (approved):

NotAShelf:
LGTM, merging once the CI passes.

NotAShelf:
Thank you :)

PR #778, opened 2025-04-02
fix: don't reference snacks in oil config

ckoehler:
I think that's a bug...


NotAShelf (approved):

NotAShelf:
Bad copy paste, thank you for catching :)

PR #781, opened 2025-04-03
language/nix: add treesitter queries for nvf idioms

horriblename:
todo: unsure if the current approach works robustly - our query has the same level of priority as nvim-treesitter in terms of rtp ordering, does this guarantee that ours will be loaded along with nvim-ts's or do we need to prepend ours in rtp?

### DIFF
--- a/flake/legacyPackages.nix
+++ b/flake/legacyPackages.nix
@@ -17,6 +17,9 @@
           # Build nil from source to get most recent
           # features as they are added.
           nil = inputs'.nil.packages.default;
+          nvf-queries = final.callPackage ./legacyPackages/nvf-queries.nix {
+            version = self.shortRev or self.dirtyShortRev or "dirty";
+          };
           blink-cmp = let
             pin = self.pins.blink-cmp;
           in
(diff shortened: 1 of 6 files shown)


snoweuph:
this brings back up the discussion whether we include queries in nvf or not.

If it were me, I would prefer having an API like `vim.queries.<name>.injections`

**examples:**

```nix
{
    cfg.gotmpl.injection = "html";
    vim.queries.gotmpl.injections = mkSCMInline ''
      ;; extends
      ((text) @injection.content
	      (#set! injection.language "${cfg.gotmpl.injection}")
	      (#set! injection.combined)
      )
    '';
}
```


```nix
{
    cfg.tera.injection = "markdown";
    vim.queries.gotmpl.injections = mkSCMInline ''
      ;; extends
      ((content) @injection.content
	      (#set! injection.language "${cfg.tera.injection}")
	      (#set! injection.combined)
      )
    '';
}
```

NotAShelf:
What would be the purpose of `mkSCMInline`? SCM is just a string in the eyes of Nix, unlike Lua which gets interpolated.

snoweuph:
> What would be the purpose of `mkSCMInline`? SCM is just a string in the eyes of Nix, unlike Lua which gets interpolated.

was just a thinking. then just make it a string, and when adding it, add lua and scm injections to NVF so I dont have to add lang comments.

Issue #777, opened 2025-04-02
formatOnSave doesnt work on .tsx files


cramt:
### ⚠️ Please verify that this bug has NOT been reported before.

- [x] I checked all existing issues and didn't find a similar issue

### Description

after the change to using conform as the formatting backend, formatOnSave doesnt seem to work for .tsx files, even though the typescript langage server is correctly enabled for those files

### 👟 Reproduction steps

```
vim.lsp.formatOnSave = true
vim.languages.ts.enable = true
```
open a .tsx file, fuck up the formatting a  bit, `:s`, see no formatting happen
important it does work for .ts files

### 👀 Expected behavior

formatting to happen

### 😓 Actual Behavior

nothing happens

### 💻 Metadata

nix-shell -p nix-info --run "nix-info -m"  - system: `"x86_64-linux"`  - host os: `Linux 6.14.0-zen1, NixOS, 25.05 (Warbler), 25.05.20250331.77b584d`  - multi-user?: `yes`  - sandbox: `yes`  - version: `nix-env (Nix) 2.24.13`  - nixpkgs: `/nix/store/9qd2bi98nh9yd4axm9zr16zln01w62q3-source`

### 📝 Relevant log output

```bash
nothing in the lsp logs
```

horriblename:
ts_ls apparently doesn't do any formatting (according to `vim.lsp.client.supports_method`), we just missed tsx in conform's config

PR #789, opened 2025-04-04
language/ts: add conform and lint rules for tsx

horriblename:
fixes: #777


NotAShelf (approved):

NotAShelf:
LGTM, but I feel like a changelog entry would be in order just to communicate the action.

PR #575, opened 2025-01-22
languages/python: add ruff as lsp

QuiNzX:
In this PR I've added ruff as lsp.

Breaking change: 
What has changed is that before ``cfg.lsp.server`` expected a single value but now it expects a list.
So that ruff can work in conjuction with basedpyright(or any other lsp) which is recommended according to the [ruff docs](https://docs.astral.sh/ruff/faq/#how-does-ruff-compare-to-mypy-or-pyright-or-pyre).

I might going to add more granular options for the lsp if you like unless you have something else in mind that will do so. [ruff lsp](https://docs.astral.sh/ruff/editors/setup/#neovim)
This is my first nix pr so let me know if I made any mistakes!

EDIT: because of the ``toList`` function the breaking change isn't the case anymore for ``cfg.lsp.server ``.
Let me know if there is a better approach to it! 
However the breaking change is still in the case of ``python.lsp.package`` which now requires each lsp server specified as an attribute set . On the bright side you can now set arguments for each lsp.
Like so: 
```nix
package = {
  ruff = [lib.getExe pkgs.ruff "server"];
  basedpyright = [lib.getExe pkgs.basedpyright];
};
```

- [x] I have updated the [changelog] as per my changes
- [x] I have tested, and self-reviewed my code
- Style and consistency
  - [x] I ran **Alejandra** to format my code (`nix fmt`)
  - [x] My code conforms to the [editorconfig] configuration of the project
  - [ ] My changes are consistent with the rest of the codebase
- If new changes are particularly complex:
  - [ ] My code includes comments in particularly complex areas
  - [ ] I have added a section in the manual
  - [ ] _(For breaking changes)_ I have included a migration guide
- Package(s) built:
  - [x] `.#nix` (default package)
  - [ ] `.#maximal`
  - [x] `.#docs-html` (manual, must build)
- Tested on platform(s)
  - [x] `x86_64-linux`
  - [ ] `aarch64-linux`
  - [ ] `x86_64-darwin`
  - [ ] `aarch64-darwin`

PR #801, opened 2025-04-05
languages: expose low-level API; convert existing modules

NotAShelf:
Based on the feedback collected from https://github.com/NotAShelf/nvf/discussions/748, and other online spaces, this pull request exposes a low-level `vim.lsp.servers` API that will be consumed internally by language modules, in a way that allows no-overhead LSP setups for both this project and end-user setups. This is, in short, the long-awaited language moduler refactor.

The new API can also be used to override LSP options set by language modules, on top of setting up LSPs not supported directly by nvf. In addition, the language modules now take a *list* of servers to allow supporting multiple servers with ease (relevant: #575)

## Modules not converted:

- [ ] Nix
- [ ] Scala 
- [ ] Rust

## Breaking Changes

1. `vim.languages.rust.crates.enable` has been renamed to `vim.languages.rust.extensions.crates-nvim.enable`

### DIFF
--- a/lib/languages.nix
+++ b/lib/languages.nix
@@ -1,11 +1,11 @@
 {lib}: let
   inherit (builtins) isString getAttr;
   inherit (lib.options) mkOption;
-  inherit (lib.types) listOf bool str submodule;
-  inherit (lib.generators) mkLuaInline;
+  inherit (lib.types) listOf bool str submodule attrsOf anything either nullOr;
   inherit (lib.nvim.attrsets) mapListToAttrs;
   inherit (lib.nvim.types) luaInline;
 in {
+  # TODO: remove
   diagnosticsToLua = {
     lang,
     config,
@@ -35,6 +35,7 @@ in {
     };
 
   lspOptions = submodule {
+    freeformType = attrsOf anything;
     options = {
       enable = mkOption {
         type = bool;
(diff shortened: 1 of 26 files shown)


thamenato:
@NotAShelf I was wondering about submitting a PR adding `setupOpts` to Python's LSP to allow changing the default `settings` - I noticed some other plugins/modules do that and thought that'd be the way to do it here as well (I don't know if there's already a way).

In any case, should I wait until you get this PR merged before working on it, or start the work based on this branch?

Tks!

NotAShelf:
> @NotAShelf I was wondering about submitting a PR adding `setupOpts` to Python's LSP to allow changing the default `settings` - I noticed some other plugins/modules do that and thought that'd be the way to do it here as well (I don't know if there's already a way).
> 
> In any case, should I wait until you get this PR merged before working on it, or start the work based on this branch?
> 
> Tks!

Sorry for not replying earlier @thamenato , I completely missed your comment somehow. `setupOpts` is not really intended for language modules, as they do not have a direct equivalent like `.setup()`. The new API that we're working on top of provides all fields, and I'd encourage you to use it instead as `vim.lsp.servers` or the new `languages.*.lsp.settings` option.

sjcobb2022:
Hi @NotAShelf,

I've started reviewing the details of this ticket, and I wanted to clarify a few points regarding the current state of the core API. I had some questions about the API design and configuration intent that I'd appreciate your input on.

#### Questions on lsp.settings

Could you clarify where the `languages.*.lsp.settings` mentioned in your previous comment is defined? I'm not seeing it in the current codebase—wondering if it's missing or located elsewhere.

#### API Design Clarifications

Is the intention of this ticket to define a fixed set of supported LSPs that users cannot customize?

From the current implementation, it appears the user cannot define their own server. If I'm interpreting the design correctly, the envisioned usage looks something like:

```nix
languages.rust = {
  enable = true;
  server = ["rust-analyzer" "anotherLSP"];
};
```

While I understand the intent to simplify configuration, this approach seems quite restrictive. Do you foresee the ability to configure LSPs in a more flexible way in the future?

When I initially started exploring this ticket, I had imagined something closer to:

```nix
languages.*.server = mkOption {
  type = listOf (attrNames lspOptions);
};
```

I think a compromise between the two approaches could also work:

```nix
type = listOf oneOf [
    (attrNames lspOptions)
    (attrNames servers)
 ];
```

That said, I also understand this might not align with the core principles of nvf configuration. Would appreciate your thoughts on whether more flexibility is in scope or if the current direction is intentional.

Thanks!

horriblename:
this is superseded by #860

the new implementation uses neovim 0.11's `vim.lsp.config`, and is as flexible as it gets

PR #804, opened 2025-04-06
utility/telescope: fix ignore patterns

ckoehler:
".git" is a regex, so because of the '.' it matches more than just the .git directory, so escape the dot to make it literal. Here, that's done with a '%'.


NotAShelf:
Looks good, could you add an entry to the changelog just to make sure users are aware of this change?

ckoehler:
Good call. Done.

NotAShelf (approved):

NotAShelf:
Thank you, merging once the CI passes.

PR #831, opened 2025-04-14
flake: bump inputs

Gerg-L:
closes: #815


Gerg-L:
awesome a check was failing because of something i didn't do (probably a alejandra update)

PR #841, opened 2025-04-18
languages/python: change DAP name from "python" to "debugpy"

howird:
- Most dap configurations expect `"debugpy"` as the adapter name, not `"python"`.
- This causes breakages in the very common case in many python repositories which have their debug config set up in `.vscode/launch.json` with `type: "debugpy"` and not `type: "python"`.
- Most people with vscode will be configuring their `launch.json` as shown [here](https://code.visualstudio.com/docs/python/debugging#_set-configuration-options).
- nvim-dap will automatically use the configuration in `.vscode/launch.json` if it exists, thus to avoid errors for nvf users that are working on repositories containing this file, we should use the same adapter as well.
- While `"python"` makes more sense as it is more general, there are no other commonly used python daps so I think renaming it to this specific implementation still makes sense.

### DIFF
--- a/docs/release-notes/rl-0.8.md
+++ b/docs/release-notes/rl-0.8.md
@@ -290,8 +290,10 @@
 
 - `eslint_d` now checks for configuration files to load.
 - Fix an error where `eslint_d` fails to load.
-- Add required files support for linters under `vim.diagnostics.nvim-lint.linters.*.required_files`.
-- Add global function `nvf_lint` under `vim.diagnostics.nvim-lint.lint_function`.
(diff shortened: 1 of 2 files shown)


NotAShelf:
Looks good to me overall. If you could just add an entry to the changelog about the rename (since it is *technically* a breaking change) we'll b golden.

howird:
Thanks, I've update the release notes.

NotAShelf (approved):

NotAShelf:
Thank you :)

PR #842, opened 2025-04-19
languages/go: default disable gofmt if lsp is enabled

horriblename:
The default go formatter is the exact same as go LSP's formatter except slower, this PR disables
gofmt if LSP is enabled

### DIFF
--- a/modules/plugins/languages/go.nix
+++ b/modules/plugins/languages/go.nix
@@ -5,7 +5,7 @@
   ...
 }: let
   inherit (builtins) attrNames;
-  inherit (lib.options) mkEnableOption mkOption;
+  inherit (lib.options) mkEnableOption mkOption literalMD;
   inherit (lib.modules) mkIf mkMerge;
   inherit (lib.meta) getExe;
   inherit (lib.lists) isList;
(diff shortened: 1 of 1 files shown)


modules/plugins/languages/go.nix
```diff
@@ -84,7 +84,14 @@ in {
...
 
     format = {
-      enable = mkEnableOption "Go formatting" // {default = config.vim.languages.enableFormat;};
+      enable =
+        mkEnableOption "Go formatting"
+        // {
+          default = !cfg.lsp.enable && config.vim.languages.enableFormat;
+          defaultText = literalMD ''
+            disabled if Go LSP is enabled, otherwise follows {option}`vim.languages.enableFormat`
+          '';
```
NotAShelf, line 93: What's `literalMD` for here? MySt-like roles should render with out I believe?

horriblename: nixos manual told me to use literalMD so I did: https://nixos.org/manual/nixos/stable/#sec-option-declarations

my understanding is text without literalMD should be rendered verbatim, but nobody does that because... nobody uses it that way?

I  don't rly care either way just let me know which you want

NotAShelf: No I was just curious, no strict preference. Just gotta make sure that it renders correctly on the documentation.

NotAShelf (approved):

PR #851, opened 2025-04-23
languages/python: fix python dap configuration regression

howird:
- [this PR](https://github.com/NotAShelf/nvf/pull/841) changed the dap adaptor name for python from `"python"` to `"debugpy"` to conform with the common case where many `.vscode/launch.json`s will reference `"debugpy"`
- Unfortunately, I introduced a regression by mistakenly changing the dap configuration name to `debugpy` as well. While the dap adaptor name refers to the adaptor type, the dap configuration name refers to the language so I reverted that one line
- I initially missed this error since I only tested the PR in a repo containing a `launch.json` so it overrode the default config, but it was brought up in [this issue](https://github.com/NotAShelf/nvf/issues/850)
- This PR has been tested __both__ in a repo containing a `launch.json` and without


NotAShelf:
Thank you :)

PR #860, opened 2025-04-25
neovim/lsp: expose low-level API

NotAShelf:
- **lib/languages: add `lspOptions` submodule type for freeform LSP config**
- **lib/languages: add `root_marker` to `lspOptions` type**
- **neovim/lsp: init**
- **lib/languages: add explicit enable to LSP submodule**
- **lib/lua: deprecate old Lua helpers superseded by `toLuaObject`**

This exposes a low(er) level API for configuring LSPs _directly_ with the new
options provided by Neovim 0.11. The idea is to allow configuring LSPs without
relying on the inclusion of language modules into nvf, or allowing language
modules in user configurations.

The language modules still use the old API, and will be migrated in due time.
Changes here are split off to avoid withholding important additions to the
module system until all language modules are converted. Language modules, of
course, will still need to be converted to this new API, after which lspconfig
shall be made into a regular plugin module instead of its current, specially
treated "variant."

Related: https://github.com/NotAShelf/nvf/discussions/748

### DIFF
--- a/lib/languages.nix
+++ b/lib/languages.nix
@@ -1,9 +1,10 @@
-# From home-manager: https://github.com/nix-community/home-manager/blob/master/modules/lib/booleans.nix
 {lib}: let
   inherit (builtins) isString getAttr;
   inherit (lib.options) mkOption;
-  inherit (lib.types) bool;
+  inherit (lib.types) listOf bool str submodule attrsOf anything;
+  inherit (lib.generators) mkLuaInline;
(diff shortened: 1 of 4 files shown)


horriblename (approved):

jeckhart:
Note: this breaks if you're not on neovim 0.11.0+ as vim.lsp.config wasn't introduced until then.

I found it as I'm on nix-stable-24.11 until 25.05 comes out, and that has neovim v0.10.2 still.

PR #884, opened 2025-05-05
cmp: use normal priority for default of `sources`

### LINKED ISSUE #839: Path error after updating flake.lock
### ⚠️ Please verify that this bug has NOT been reported before.

- [x] I checked all existing issues and didn't find a similar issue

### Description

After updating my flake.lock, my configuration have issues dealing with paths.
For example the telescope explorer can't create new files/directories and gives the error message:

`[file_browser.to_absolute_path] Given path ':%:p:h' doesn't exist`

And nvim-cmp/blink-cmp doesn't autocomplete paths anymore. If i revert to the old lock file, i dont have these issues. My previous lock file is from March 12th

### 👟 Reproduction steps

use telescope file explorer or nvim-cmp path autocompletion

### 👀 Expected behavior

Be able to get autocompletion on paths, and create files/folders with the explorer

### 😓 Actual Behavior

Either getting error message, or no completion.

### 💻 Metadata

 - system: `"x86_64-linux"`  - host os: `Linux 6.14.2, NixOS, 24.11 (Vicuna), 24.11.20250412.26d499f`  - multi-user?: `yes`  - sandbox: `yes`  - version: `nix-env (Nix) 2.24.14`  - nixpkgs: `/nix/store/hzaj4d6ari2wq2cbg1j60n9zw42gnshy-source`

### 📝 Relevant log output

```bash
-
```

horriblename:
fixes #839

After this change, user configs are "appended" to default ones
instead of overriding them

### DIFF
--- a/modules/plugins/completion/nvim-cmp/config.nix
+++ b/modules/plugins/completion/nvim-cmp/config.nix
@@ -60,6 +60,11 @@ in {
         enableSharedCmpSources = true;
 
         nvim-cmp = {
+          sources = {
+            nvim-cmp = null;
+            buffer = "[Buffer]";
+            path = "[Path]";
+          };
           sourcePlugins = ["cmp-buffer" "cmp-path"];
 
           setupOpts = {
--- a/modules/plugins/completion/nvim-cmp/nvim-cmp.nix
+++ b/modules/plugins/completion/nvim-cmp/nvim-cmp.nix
@@ -98,11 +98,15 @@ in {
 
     sources = mkOption {
       type = attrsOf (nullOr str);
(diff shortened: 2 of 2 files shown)


horriblename:
based on how long it took for #839 to be resolved I would disagree, people expect the default value to be extended rather than overwritten

alternatively, we could write it down somewhere how listOf/attrsOf options are expected to work

NotAShelf:
Okay, let's get this fix out of the way and revisit the underlying issue again in the future. I think explicit override is a sensitive matter, and I'd like to discuss it when I have less on my plate.

PR #885, opened 2025-05-05
docs/hacking: add notes about listOf default priority

horriblename:
I finished writing this before realizing I should have checked how `listOf` is
used in other codebases first.

I did a search for `listOf` in nixpkgs codebase, and none of the results on
the first page make use of `listOf` in a way that the default values should be
extended. (in fact, almost all of them set `default = []`)

With that in mind, I'm not sure if what I wrote in this PR is good practice or
just a hack that would just makes things more complicated down the line


horriblename:
dumb idea

PR #892, opened 2025-05-09
feat: add treesitter textobjects

ckoehler:
Not sure if there's a better way to make this load before mini.ai and after TS. Nothing seems to break even if mini.ai is disabled.
I also don't use any default config for the actual plugin, just the objects it provides with my mini.ai user config.

### DIFF
--- a/modules/plugins/treesitter/ts-textobjects/textobjects.nix
+++ b/modules/plugins/treesitter/ts-textobjects/textobjects.nix
@@ -1,0 +1,21 @@
+{lib, ...}: let
+  inherit (lib.options) mkEnableOption;
+  inherit (lib.nvim.types) mkPluginSetupOption;
+in {
+  options.vim.treesitter.textobjects = {
+    enable = mkEnableOption "Treesitter textobjects";
+    setupOpts =
+      mkPluginSetupOption "treesitter-textobjects" {}
+      // {
+        example = {
+          select = {
+            enable = true;
+            lookahead = true;
+            keymaps = {
+              af = "@function.outer";
+            };
+          };
+        };
+      };
+  };
+}
--- a/docs/release-notes/rl-0.8.md
+++ b/docs/release-notes/rl-0.8.md
@@ -329,11 +329,13 @@
 
 [flash.nvim]: https://github.com/folke/flash.nvim
 [gitlinker.nvim]: https://github.com/linrongbin16/gitlinker.nvim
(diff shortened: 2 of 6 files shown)


horriblename:
I don't understand why mini.ai has to come first? from their README I don't see any dependency between the two?

horriblename:
did a bit more reading in their docs, still no mention of needing to load one before the other

ckoehler:
> did a bit more reading in their docs, still no mention of needing to load one before the other

See https://github.com/echasnovski/mini.nvim/blob/4e873f36552ac7c15a824f36783c629493f78d0c/doc/mini-ai.txt#L725. If you don't, that query file won't be available and the motions just won't work.

horriblename:
the query files will always be there. the only way they could be missing is if you lazy-load TS-textobject, which isn't happening here

ckoehler:
> the query files will always be there. the only way they could be missing is if you lazy-load TS-textobject, which isn't happening here

You're right. I did have it lazy-loading, which required this. But then I changed it. Thanks for pushing back. This makes it easier, too :)

modules/plugins/treesitter/ts-textobjects/textobjects.nix
```diff
@@ -0,0 +1,9 @@
+{lib, ...}: let
+  inherit (lib.options) mkEnableOption;
+  inherit (lib.nvim.types) mkPluginSetupOption;
+in {
+  options.vim.treesitter.textobjects = {
+    enable = mkEnableOption "Treesitter textobjects";
+    setupOpts = mkPluginSetupOption "treesitter-textobjects" {};
```
horriblename, line 7: an example would be appreciated
```suggestion
    setupOpts = mkPluginSetupOption "treesitter-textobjects" {} // {
      example = {
        select = {
          enable = true,
          lookahead = true,
          keymaps = {
            ["af"] = "@function.outer",
          },
      };
    };
  };
```

would hopefully make it clear enough that the whole thing goes under `textobjects` in the generated code

horriblename (changes requested):

modules/plugins/treesitter/ts-textobjects/textobjects.nix
ckoehler: Done, thanks!

horriblename (approved):
LGTM, thanks!

NotAShelf (commented):
LGTM

NotAShelf:
Thanks :)

PR #894, opened 2025-05-10
add ruby lsp as ruby server option

cramt:
added shopify's new lsp for ruby called `ruby-lsp` as a server option along side solargraph

Also updated the solargraph config to match all the other, with the fact that it can either be a package or list of strings


NotAShelf:
Sorry for the delay, could you add an entry in the release notes please? I'll merge it afterwards.

cramt:
no problem, it was the weekend after all, have added release notes now

Soliprem (approved):

PR #900, opened 2025-05-15
fix: disable Conform for Rust

ckoehler:
I don't think we want Conform to also do autoformatting. I am running into issues where it doesn't respect the `rustfmt.toml` file and will happily format according to the 2015 edition. The builtin `RustFmt` command also works and can be set up to autosave, but `rustaceanvim` includes `rust-analyzer` which will run `rustfmt` as well. 

All this to say, autoformatting still works without Conform and also doesn't break stuff.

### DIFF
--- a/docs/release-notes/rl-0.8.md
+++ b/docs/release-notes/rl-0.8.md
@@ -346,6 +346,8 @@
 - Fix default telescope ignore list entry for '.git/' to properly match
 - Add [gitlinker.nvim] plugin to `vim.git.gitlinker-nvim`
 - Add [nvim-treesitter-textobjects] plugin to `vim.treesitter.textobjects`
+- Default to disabling Conform for Rust if rust-analyzer is used
+  - To force using Conform, set `languages.rust.format.enable = true`.
 
 [rrvsh](https://github.com/rrvsh):
 
--- a/modules/plugins/languages/rust.nix
+++ b/modules/plugins/languages/rust.nix
@@ -7,7 +7,7 @@
   inherit (builtins) attrNames;
   inherit (lib.meta) getExe;
   inherit (lib.modules) mkIf mkMerge;
-  inherit (lib.options) mkOption mkEnableOption;
+  inherit (lib.options) mkOption mkEnableOption literalMD;
   inherit (lib.strings) optionalString;
   inherit (lib.trivial) boolToString;
   inherit (lib.lists) isList;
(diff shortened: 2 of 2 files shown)


ckoehler:
Ah changelog, standby for that.

horriblename:
don't delete the formatter config, do this instead: https://github.com/NotAShelf/nvf/blob/74587e254ac7f8f7a7eca58aede14badde20dee6/modules/plugins/languages/go.nix#L90

ckoehler:
> don't delete the formatter config, do this instead:
> 
> https://github.com/NotAShelf/nvf/blob/74587e254ac7f8f7a7eca58aede14badde20dee6/modules/plugins/languages/go.nix#L90

Done.

horriblename:
uh, I meant change the deafult value of the _option_ (+ add defaultText), leave the config section alone

https://github.com/NotAShelf/nvf/commit/b7627523333b1ac64111e0f030934f9d17dbb063#diff-ceba2b90515a5fc940fc0dc9d5677f1c27555c890ecd2677ca854effc05164a6R86-R94

docs/release-notes/rl-0.8.md
```diff
@@ -346,6 +346,7 @@
 - Fix default telescope ignore list entry for '.git/' to properly match
 - Add [gitlinker.nvim] plugin to `vim.git.gitlinker-nvim`
 - Add [nvim-treesitter-textobjects] plugin to `vim.treesitter.textobjects`
+- Disable Conform for Rust if rust-analyzer is enabled
```
NotAShelf, line 349: Can we have a note on that this is the default behaviour, and that it can be overriden if the user desires. A nested point detailing why and instructing how to "fix" this behaviour would be appreciated.

modules/plugins/languages/rust.nix
```diff
@@ -68,7 +68,14 @@ in {
     };
 
     format = {
-      enable = mkEnableOption "Rust formatting" // {default = config.vim.languages.enableFormat;};
+      enable =
+        mkEnableOption "Rust formatting"
+        // {
+          default = !cfg.lsp.enable && config.vim.languages.enableFormat;
+          defaultText = literalMD ''
+            disabled if Rust LSP is enabled, otherwise follows {option}`vim.languages.enableFormat`
```
NotAShelf, line 76:
```suggestion
            Disabled if Rust LSP is enabled, otherwise follows {option}`vim.languages.enableFormat`
```

NotAShelf (changes requested):
Two quick nits, looks good to merge afterwards.

docs/release-notes/rl-0.8.md
ckoehler: Done.

horriblename (approved):

horriblename:
LGTM thanks!

NotAShelf (approved):

PR #906, opened 2025-05-17
update mnw, fix noBuildPlug

Gerg-L:


modules/wrapper/build/config.nix
```diff
@@ -7,23 +7,20 @@
...
     pin = getPin pname;
-    version = pin.revision or "dirty";
-  in {
-    # vim.lazy.plugins relies on pname, so we only set that here
-    # version isn't needed for anything, but inherit it anyway for correctness
-    inherit pname version;
-    outPath = path {
-      name = "${pname}-0-unstable-${version}";
-      path = pin.outPath;
+    version = builtins.substring 0 8 pin.revision;
```
NotAShelf, line 16: Curiousity more than a nit, but what if revision is less than 8 characters or doesn't exist? Wouldn't this cause a nasty eval failure then?

Gerg-L: Yeah, if a the npin isn't a git repo

NotAShelf:
wtf is happening to my CI

PR #913, opened 2025-05-20
Draft: Language overhall 

sjcobb2022:
Hi,

This PR begins refactoring the language module to use `vim.lsp.config` instead of `lspconfig`.

The primary motivation is to explore new design patterns for how the API could be structured. Currently, this is only implemented for the `nix` language.

### Design 1

```nix
lspConfigurations =
  mapAttrsToList (
    name: value:
      ''vim.lsp.config["${name}"] = vim.tbl_deep_extend("force", lspconfig.${name}, ${toLuaObject value})''
  )
  (filterAttrs (n: _: n != "*") cfg.servers);

globalConfiguration =
  mapAttrsToList (
    name: value:
      ''vim.lsp.config["${name}"] = ${toLuaObject value}''
  )
  (filterAttrs (n: _: n == "*") cfg.servers);
```

```nix
vim.luaConfigRC.lsp-servers = entryAfter ["lspconfig"] ''
  ${concatLines globalConfiguration}
  ${concatLines lspConfigurations}
  vim.lsp.enable(${toLuaObject (filter (name: name != "*") (attrNames enabledServers))});
'';
```

### Design 2

```nix
# Configure each language separately
vim.luaConfigRC.lspconfig-${languageName} = let
  configuredLspConfigs = map (
    name:
      ''vim.lsp.config["${name}"] = vim.tbl_deep_extend("force", lspconfig.${name}, vim.lsp.config["${name}"])''
  ) (attrNames cfg.lsp.servers);
in
  entryAfter ["lspconfig" "lsp-servers"] ''
    ${concatLines configuredLspConfigs}
  '';
```

### Design 3

Simplify the API entirely by removing support for custom LSP configs in individual language modules. Instead, all configuration would be done through `vim.lsp.servers`, allowing Nix to handle merging automatically.

### DIFF
--- a/lib/languages.nix
+++ b/lib/languages.nix
@@ -1,9 +1,13 @@
 {lib}: let
   inherit (builtins) isString getAttr;
   inherit (lib.options) mkOption;
-  inherit (lib.types) listOf bool str submodule attrsOf anything either nullOr;
+  inherit (lib.strings) concatStringsSep;
+  inherit (lib.types) listOf bool str submodule attrsOf anything either nullOr oneOf enum;
+  inherit (lib.attrsets) attrNames;
   inherit (lib.nvim.attrsets) mapListToAttrs;
   inherit (lib.nvim.types) luaInline;
+  inherit (lib.lists) isList;
+  inherit (lib) genAttrs recursiveUpdate;
 in {
   # TODO: remove
   diagnosticsToLua = {
@@ -33,6 +37,18 @@ in {
       type = bool;
       description = "Turn on ${desc} for enabled languages by default";
     };
+
+  # resolveLspOptions
+  # servers: AttrsOf lspOptions
+  # selected: AttrsOf lspOptions | List of string keys from servers
+  # Returns: AttrsOf lspOptions
+  resolveLspOptions = {
+    servers,
+    selected,
+  }:
+    if isList selected
+    then genAttrs selected (name: servers.${name})
+    else selected;
 
   lspOptions = submodule {
     freeformType = attrsOf anything;
@@ -77,4 +93,27 @@ in {
       };
     };
   };
+
+  mkLspOption = {servers, ...} @ args: let
+    serverNames = attrNames servers;
+    defaultAttrs = {
+      type = oneOf [
+        (attrsOf lib.nvim.languages.lspOptions)
+        (listOf (enum serverNames))
+      ];
+      description = ''
+        Either a full set of selected LSP options as an attribute set,
+        or a list of server names from: ${concatStringsSep ", " serverNames}.
+      '';
+      default = {};
+      example = {
+        clangd = {
+          filetypes = ["c"];
+          root_markers = ["CMakeLists.txt"];
+        };
+      };
+    };
+    cleanedArgs = removeAttrs args ["servers"];
+  in
+    mkOption (recursiveUpdate defaultAttrs cleanedArgs);
 }
--- a/modules/neovim/init/lsp.nix
+++ b/modules/neovim/init/lsp.nix
@@ -18,11 +18,35 @@
 
   lspConfigurations =
     mapAttrsToList (
-      name: value: ''
-        vim.lsp.config["${name}"] = ${toLuaObject value}
-      ''
+      # TODO: Determine the best thing to do about merging in lspconfig
+      name: value:
+      /*
+      lua
+      */
+      ''vim.lsp.config["${name}"] = ${toLuaObject value}''
     )
     cfg.servers;
+
+  # Approach 1:
+  # Create function perhaps called mkLspConfig
+  # mkLspConfig servers;
+  # that expands to something like
+  # vim.lsp.servers = cfg.lsp.servers......
+  # vim.luaConfigRC.lspconfigMerge = entryAfter ["lsp-servers"] ''vim.lsp.config["${name}"] = vim.tbl_deep_extend("force", lspconfig.${name}, vim.lsp.config["${name}"])''
+
+
+  # Approach 2:
+  # lspConfigurations =
+  #   mapAttrsToList (
+  #     name: value: ''vim.lsp.config["${name}"] = vim.tbl_deep_extend("force", lspconfig.${name}, ${toLuaObject value})''
+  #   )
+  #   (filterAttrs (n: _: n != "*") cfg.servers);
+  # Then also need to configure global * settings
+  # globalConfiguration =
+  #   mapAttrsToList (
+  #     name: value: ''vim.lsp.config["${name}"] = ${toLuaObject value}''
+  #   )
+  #   (filterAttrs (n: _: n == "*") cfg.servers);
 
   enabledServers = filterAttrs (_: u: u.enable) cfg.servers;
 in {
@@ -81,13 +105,18 @@ in {
     }
 
     (mkIf (cfg.servers != {}) {
-      vim.luaConfigRC.lsp-servers = entryAnywhere ''
-        -- Individual LSP configurations managed by nvf.
-        ${concatLines lspConfigurations}
+      vim.luaConfigRC.lsp-servers =
+        entryAnywhere
+         # Or entryAfter ["lspconfig"] if we go with the second approach
+        /*
+        lua
+        */
+        ''
+          ${concatLines lspConfigurations}
 
-        -- Enable configured LSPs explicitly
-        vim.lsp.enable(${toLuaObject (filter (name: name != "*") (attrNames enabledServers))})
-      '';
+          -- Enable configured LSPs explicitly
+          vim.lsp.enable(${toLuaObject (filter (name: name != "*") (attrNames enabledServers))});
+        '';
     })
   ];
 }
--- a/modules/plugins/languages/nix.nix
+++ b/modules/plugins/languages/nix.nix
@@ -2,99 +2,54 @@
   config,
   pkgs,
   lib,
-  inputs,
   ...
 }: let
   inherit (builtins) attrNames;
   inherit (lib) concatStringsSep;
+  inherit (lib.generators) mkLuaInline;
   inherit (lib.meta) getExe;
   inherit (lib.options) mkEnableOption mkOption;
   inherit (lib.modules) mkIf mkMerge;
   inherit (lib.lists) isList;
-  inherit (lib.strings) optionalString;
-  inherit (lib.types) anything attrsOf enum either listOf nullOr package str oneOf;
+  inherit (lib.types) enum package;
   inherit (lib.nvim.types) mkGrammarOption diagnostics;
-  inherit (lib.nvim.lua) expToLua toLuaObject;
-  inherit (lib.nvim.languages) lspOptions;
+  inherit (lib.nvim.lua) expToLua;
+  inherit (lib.nvim.languages) resolveLspOptions mkLspOption;
 
   cfg = config.vim.languages.nix;
 
-  useFormat = "on_attach = default_on_attach";
-  noFormat = "on_attach = attach_keymaps";
-
-  defaultServer = "nil";
   packageToCmd = package: defaultCmd:
     if isList package
     then expToLua package
     else ''{"${package}/bin/${defaultCmd}"}'';
+
+  formattingCmd = lib.mkIf (cfg.format.enable && cfg.lsp.enable) {
+    formatting = lib.mkMerge [
+      (lib.mkIf (cfg.format.type == "alejandra") {
+        command = ["${cfg.format.package}/bin/alejandra" "--quiet"];
+      })
+      (lib.mkIf (cfg.format.type == "nixfmt") {
+        command = ["${cfg.format.package}/bin/nixfmt"];
+      })
+    ];
+  };
+
+  defaultServers = ["nil_ls"]
   servers = {
-    nil = {
-      package = inputs.nil.packages.${pkgs.stdenv.system}.nil;
-      internalFormatter = true;
-      # lspConfig = ''
-        # lspconfig.nil_ls.setup{
-        #   capabilities = capabilities,
-        # ${
-        #   if cfg.format.enable
-        #   then useFormat
-        #   else noFormat
-        # },
-        #   cmd = ${packageToCmd cfg.lsp.package "nil"},
-        # ${optionalString cfg.format.enable ''
-        #   settings = {
-        #     ["nil"] = {
-        #   ${optionalString (cfg.format.type == "alejandra")
-        #     ''
-        #       formatting = {
-        #         command = {"${cfg.format.package}/bin/alejandra", "--quiet"},
-        #       },
-        #     ''}
-        #   ${optionalString (cfg.format.type == "nixfmt")
-        #     ''
-        #       formatting = {
-        #         command = {"${cfg.format.package}/bin/nixfmt"},
-        #       },
-        #     ''}
-        #     },
-        #   },
-        # ''}
-        # }
-      # '';
+    nil_ls = {
+      enable = true;
+      cmd = ["${pkgs.nil}/bin/nil"];
+      settings = {
+        nil = formattingCmd;
+      };
     };
 
     nixd = {
-      package = pkgs.nixd;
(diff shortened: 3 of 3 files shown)


horriblename:
Note that we're delaying the language overhaul until after 25.05 is released, to prevent further breakage to users on nixpkgs-stable

We have talked about this before and the current plan is to just remove customization from language modules (design 3), though it's not set in stone yet.

NotAShelf:
25.05 is coming soon, so it is best that we discuss this soon. Keep in mind that

1. language modules should not rely on lspconfig
2. API breakage should be as minimal as possible
3. we want full customizability to the degree that it is possible

I'm a little too busy and too occupied right now to think about this *right now*, but I am also glad there is initiative. If all goes well I can provide a review and some guidance this friday.

sjcobb2022:
Discussion point for Friday:

> language modules should not rely on lspconfig

Does this means that we should never mention lspconfig in `modules/nvim/init/lsp.nix`? Or does it mean that it shouldn't be in the language module (which it already is not in this branch)

If it is the latter, then I think we already have a pretty good solution for this:

```nix
''vim.lsp.config["${name}"] = vim.tbl_deep_extend("force", lspconfig.${name}, vim.lsp.config["${name}"])''
```

This implies 2 things:
- the name of the lsp server should match that in lspconfig
- lspconfig.${name} should always exist if it is present.
    - could be easily solved with a little bit of inline lua logic.

> Brain dump comment
> The matching names is a little hard as we cannot verify this in our code. Perhaps a janky, and perhaps uncoventional, solution would be to rely on lspconfig as a non-flake input, and map the files in the git repo to a list of supported lsp names.

horriblename:
> and we rely more heavily on lib.getExe to ensure that we are running the correct packages.

we can't, `clang-tools` doesn't have a mainProgram, neither does haskell LSP. (for good reason)

if we were to add a "package" option, it needs to accept string path as well

> This raises the issues however that our own configuration could cause code breaks with newer LSP package versions

what do you mean by this?

horriblename:
on the point of name conflicts with lspconfig: the easiest solution would be to prefix "nvf_" to lsp names, I'm not sure I like the idea, but this does come with the benefit of being obviously distinct from lspconfig, therefore avoiding user confusion as to where the config comes from (realistically, this is most people)

sjcobb2022:
> if we were to add a "package" option, it needs to accept string path as well

That could work

> what do you mean by this?

I mean if the LSP changes how the function (e.g. depracates or removes a --stdio flag or something) we would break the 

> on the point of name conflicts with lspconfig: the easiest solution would be to prefix "nvf_" to lsp names, I'm not sure I like the idea, but this does come with the benefit of being obviously distinct from lspconfig, therefore avoiding user confusion as to where the config comes from

If we were to do this, we would need a separate mechanism to merge lspconfig configurations into our own configurations. At the moment, by matching the lspconfig names we get the configs merged in automatically.


There are 2 approaches that we can go for imo:
- Configure everything ourselves and no longer rely on lspconfig at all (most likely end up basing our own config on lspconfig)
- Still rely on lspconfig and maintain naming conventions with lspconfig and benefit from the auto-merging of configs.

sjcobb2022:
I believe my previous comment is slightly misleading

e.g. "The issue with matching lspconfig lsp names to our own is still present, however this could be managed by hand."

The issue is that we have no "proper" mapping. Ideally I think we should maintain a mapping of language lsp configs to java lsp configs

horriblename:
> I mean if the LSP changes how the function (e.g. depracates or removes a --stdio flag or something) we would break the
Not really much of a concern, if upstream changes something we simply update our option





> There are 2 approaches that we can go for imo:
> 
>     * Configure everything ourselves and no longer rely on lspconfig at all (most likely end up basing our own config on lspconfig)
> 
>     * Still rely on lspconfig and maintain naming conventions with lspconfig and benefit from the auto-merging of configs.

The plan is to not use lspconfig in language modules (this is what NotAShelf means by "language modules should not rely on lspconfig")

though I don't remember why we came to that decision. @NotAShelf if you could remind me why please

sjcobb2022:
This current branch already removes lspconfig totally from language modules, but still requires lspconfig to be active so that the lspconfig lua can automatically merge it's predefined configs.

e.g.

```lua
-- LSPCONFIG section
local lspconfig = require('lspconfig')  -- pulls in lspconfig opts

-- LSPSERVERS section
vim.lsp.enable(....) -- all of these now have merged in configs
```

horriblename:
and that's depending on lspconfig. What we meant is language modules should work even if lspconfig is not installed. Basically this means copying configs from lspconfig to our language module

sjcobb2022:
gotcha, that makes sense

NotAShelf:
Sorry for not returning to you earlier, I was quite tired yesterday after work.

> After some further testing, for whatever reason I got into my head that lspconfig did not automatically merge configs (it does). I believe that I previously had the wrong name for the lspconfig (nills instead of nil_ls) which caused the merged lspconfig to not be present. This means that we can keep the existing vim.lsp.servers api and init without having to change anything.

We do not need to rely on lspconfig to merge conflicts. As mentioned before, I'd like to drop our dependency on lspconfig entirely; it's a wrapper around the Neovim's own APIs to "generate" the configs, which is the exact role nvf fulfils. lspconfig adds unnecessary overhead, we can handle Nix -> Lua instead of doing Nix -> Lua (lspconfig) -> Lua (vim.lsp). I think `vim.lsp.servers` is flexible enough for what we are trying to do. We can add a conformant `.settings` or `.options` submodule to each language module that handles conversion to `vim.lsp.servers` in our module system. This would also make configuring frictionless.

> At the moment the cmds are formatted in a way that if you wanted to override a package you would need to rewrite the entire command. I don't think that should necessarily be the responsibility of the user.

> Perhaps an additional package field as well, and we rely more heavily on lib.getExe to ensure that we are running the correct packages.

One interesting idea would be to use `lib.getExe'` and specify the binary name. If the user wishes to override the package, then they would be responsible for providing an output with the same name. This is also added complexity however, so I'm open to alternatives. Maybe separating command and arguments would work.

sjcobb2022:
@horriblename in that case I don't really see a reason to split the config into separate files then (other than for code clarity).

We are already separated by language so it makes sense to leave it like that.

horriblename:
that's fine as well (for now), long term I want to find a way to reduce the code duplication in js and friends, but not necessary for this rework.

lib/languages.nix
```diff
@@ -62,7 +62,7 @@ in {
       };
 
       cmd = mkOption {
-        type = nullOr (listOf str);
+        type = nullOr (listOf (oneOf [str luaInline]));
```
horriblename: pretty sure it should be

```
nullOr (either luaInline (listOf str)) 
```

modules/plugins/languages/astro.nix
```diff
@@ -12,25 +12,46 @@
...
+                local stat = vim.loop.fs_stat(typescript_path)
+                if stat and stat.type == 'directory' then
+                  return typescript_path .. '/lib'
+                end
+              end
+              return '''
+            end
+
+            if config.init_options and config.init_options.typescript and not config.init_options.typescript.tsdk then
+              config.init_options.typescript.tsdk = get_typescript_server_path(config.root_dir)
```
horriblename: is init_pptions and init_options.typescript possibly nil here?

if guaranteed not nil, why the check in the previous line?

if not guaranteed, a table should be created first (or use vim.tbl_extend)

lib/languages.nix
sjcobb2022: I made it like this so we could do 
```nix
[
   (lib.getExe package)
   mkLuaInline "some_function()"
]
```

But I totally see that `nullOr (either luaInline (listOf str))` would be a preferred api

modules/plugins/languages/astro.nix
sjcobb2022: Copied from nvim lspconfig, will do some testing.

sjcobb2022:
@horriblename @NotAShelf After looking through the lspconfig repo more carefully, there are a lot of utility functions that we would need to replicate here.

Since nvim-lspconfig project has more visibility and contributors, it makes sense to delegate the responsibility of maintaining sane defaults to them.

If these are not your thoughts, then I will add a util lua section and duplicate nvim-lspconfig utils.

sjcobb2022:
> @horriblename @NotAShelf After looking through the lspconfig repo more carefully, there are a lot of utility functions that we would need to replicate here.
> 
> Since nvim-lspconfig project has more visibility and contributors, it makes sense to delegate the responsibility of maintaining sane defaults to them.
> 
> If these are not your thoughts, then I will add a util lua section and duplicate nvim-lspconfig utils.

Bump on this

horriblename:
I don't mind keeping lspconfig honestly

NotAShelf:
Thanks for the ping, Github decided I shouldn't get notifs on my own repo for some reason...

I'm honestly more in favor of dropping lspconfig, but if it's going to be a hassle then it can be kept. Such is Neovim...

willcl-ark:
I have removed nvim-lspconfig before (in my native Lua nvim config) and from what I remember the primary util that I felt I wanted to replicate was their more advanced `root_file`-finding logic, i.e.: https://github.com/neovim/nvim-lspconfig/blob/77d3fdfb3554632c7a3b101ded643d422de7626f/lua/lspconfig/util.lua#L27-L78

The others I was just fine without. I've since removed the root util too, and tbh been fine without it. @sjcobb2022 did you have other utils in particular you thought looked necessary?

Personally I'd also slightly lean towards preferring to remove nvim-lspconfig, relying on pass-through to the native setup. If the (nvim) native root-finding is deficient for purpose then ideally I'd hope a fix would come from upstream nvim, rather than requiring an entire new plugin for a util or two.

All that said, I'm not doing the work here; I'm just glad someone is. So thanks!

sjcobb2022:
I might just port over the util and build from there.

sjcobb2022:
Converted modules:

40 / 43 (including #962)

- [x] asm.nix
- [x] astro.nix
- [x] bash.nix
- [x] \*clang.nix
- [x] clojure.nix
- [x] \*csharp.nix
- [x] \*css.nix
- [x] \*cue.nix
- [x] dart.nix
- [x] elixir.nix
- [x] \*fsharp.nix
- [x] gleam.nix
- [x] \*go.nix
- [x] haskell.nix
- [x] hcl.nix
- [x] helm.nix
- [x] html.nix (nothing to do)
- [x] java.nix
- [x] julia.nix
- [x] kotlin.nix
- [x] lua.nix
- [x] markdown.nix
- [x] nim.nix
- [x] nix.nix
- [x] nu.nix
- [x] ocaml.nix
- [x] odin.nix
- [x] php.nix
- [x] python.nix
- [x] r.nix
- [x] ruby.nix
- [ ] rust.nix
- [ ] scala.nix
- [x] sql.nix
- [x] svelte.nix
- [x] tailwind.nix
- [x] terraform.nix
- [ ] ts.nix
- [x] typst.nix
- [x] vala.nix
- [x] wgsl.nix
- [x] yaml.nix
- [x] zig.nix

\* in #962

horriblename:
do you need help with those? to speed up the process

sjcobb2022:
If you would like to convert a few modules that would be great, but not necessary. I have just been doing these when I get a bit of time in the evenings. If you want this merged in sooner rather than later, any and all contributions would be welcome

horriblename:
great, I'll try to do a few when I get time

horriblename:
I added a few languages in #962, I'll keep updating your list as I go

sjcobb2022:
Added java and julia

sjcobb2022:
3 more to go!

These are the 3 more complicated ones imo which is why i saved them for last

horriblename:
we can skip rust and scala - those are managed by their own plugins, and while they deserve a bit of a rework themselves, they're not using `vim.lsp.config` directly, so we can save those for their own dedicated PRs.

I have the typescript one ready to go, and with this, we covered all languages. :tada: 

The remaining work needed is to add deprecation warnings/errors. I haven't decided what exactly to do with the removed options yet so I'll look into that and get back to you in a bit

thanks for the great work so far!

horriblename:
I can add in the missing deprecation warnings in #962, in the meantime, @NotAShelf this PR is ready for review

NotAShelf:
Awesome, thank you both. I'm going to review this when I'm back from work tomorrow, please remind meif I do end up forgetting.

configuration.nix
```diff
@@ -82,7 +82,8 @@ isMaximal: {
       dart.enable = false;
       ocaml.enable = false;
       elixir.enable = false;
-      haskell.enable = false;
+      haskell.enable = true;
+      hcl.enable = true;
```
NotAShelf: Please keep those disabled. `configuration.nix` dictates the servers enabled for the default packages, and Haskell toolchain is too large to be pulling for the default package.

modules/plugins/languages/asm.nix
```diff
@@ -20,12 +32,11 @@ in {
...
-
-      package = mkOption {
-        type = package;
-        default = pkgs.asm-lsp;
-        description = "asm-lsp package";
+      enable = mkEnableOption "Assembly LSP support" // {default = config.vim.lsp.enable;};
+      servers = mkOption {
+        description = "Assembly LSP server to use";
+        type = listOf (enum (attrNames servers));
+        default = defaultServers;
```
NotAShelf: Please keep the ordering as

```
2. type
2. default
3. (optional) example
4. description
```

NotAShelf (changes requested):
Looks good to me overall. I've left two comments on site, but here are two additional notes:

1. This is a massive breaking change. We are completely refactoring how language servers are configured, and I believe it should be fully documented. Please include a changelog entry as well as a highlight in the breaking changes section. The `lsp.server` options have been changed.
2. Please handle the renamed options with `mkRenamedOptionModule`. It is a horrible experience for the users otherewise.

As a sidenote/nit please order mkOption arguments as in my comment. Everything else LGTM.

horriblename:
> Please handle the renamed options with mkRenamedOptionModule. It is a horrible experience for the users otherewise.

I'm doing that in #962 keep in mind though, that `package` options can't be coerced into our `cmd` options so those are `mkRemovedOptionModule`'d.

modules/plugins/languages/asm.nix
sjcobb2022: done [e0cd56f](https://github.com/NotAShelf/nvf/pull/913/commits/e0cd56f39e218546d8258bd838754e06df64d792)

configuration.nix
sjcobb2022: done [a42d6ac](https://github.com/NotAShelf/nvf/pull/913/commits/a42d6ac69aeee101f12b99a2486e386a0fa8bd2a)

sjcobb2022:
I think waiting for @horriblename to complete #962 with the completions and then we are good to go.

Should we merge #962 into this?

NotAShelf:
As previously discussed with @horriblename, I've rebased this on v0.8--which is our "branch-out" ref. It'll be used as a staging area for our changes until we are absolutely positive that everything is stable, and then it will be merged into main. That said, thank you for everything. I'm going to merge this in the morning after a final review, but I'm positive I can handle any remaining chages myself as I have some time to spare. Once this is merged, we can rebase #962 and merge it on top. That should be *almost* everything for v0.8, with the last remaining change being to migrating to the new documentation utility.

sjcobb2022:
Awesome sauce, thanks @NotAShelf

Soliprem:
o.O there's some failing CI

Soliprem:
@NotAShelf there's build failures for docs-manpages, hyperlinks and typos

Soliprem:
the typo one is useless, however, because it's just words we need to add to the regex. The other two I'm still parsing

Soliprem:
the other two are caused by julia.lsp.servers. Honestly can't quite see what's wrong with it at first glance, though

NotAShelf:
*This* is why you don't use Python for critical utilities. Incredible error reporting.

```
#opt-vim.languages.julia.lsp.package
```

The above value was used whereas it should *probably* have been package*s*.

PR #914, opened 2025-05-21
languages/rust: add rustaceanvim dependency rustfmt

Noah765:
Rust formatting doesn't work when `rustfmt` isn't available in your `$PATH`. The `ft_rust` plugin, which provides the `:RustFmt` command and autoformatting and exposes an option called `g:rustfmt_command`. However, this option only accepts the name of an executable in `$PATH`, not the full path to the executable (see https://neovim.io/doc/user/ft_rust.html). Therefore, we must add `rustfmt` to `extraPackages`.

## Sanity Checking

- Style and consistency
- Package(s) built:
  - [ ] `.#maximal`
- Tested on platform(s)
  - [x] `x86_64-linux`
  - [ ] `aarch64-linux`
  - [ ] `x86_64-darwin`
  - [ ] `aarch64-darwin`

---


horriblename:
note that `ft_rust` comes builtin with neovim, and is not part of rustaceanvim.

formatting in language modules are typically handled by either conform.nvim or the LSP. I don't really see a reason to change that for rust.

Noah765:
My explanation of why I called `rustfmt` a dependency of `rustaceanvim`, and how that relates to `ft_rust`, was a bit lacking. I updated the description of this PR and the commit message.

Noah765:
To reiterate, `rust-analyzer` is the de facto LSP client for Rust, and Conform is disabled by default. The formatting functionality of `rust-analyzer` does not work when `rustfmt` is unavailable. Currently, this is the default behavior. Depending on packages being available in `$PATH` goes against core Nix principles.
To avoid such system-dependent behavior, I think `rustfmt` should be provided by default.
However, I agree that there should be an option to override the package, or even set it to null to explicitly depend on `rustfmt` being in `$PATH`. I considered reusing the `format.package` option, but that might cause confusion because the `package` option would affect formatting even when `format.enable` is set to false. I think adding an option, maybe to the `lsp` options, may be the best solution. Undoing https://github.com/NotAShelf/nvf/pull/900 would also work, of course. What are your thoughts?

horriblename:
ahh, I missed that part, sorry. 

so the options are

1. revert #900, manage rustfmt via conform (:RustFmt is borked)
2. add rustfmt to extraPackages (:RustFmt and rust-analyzer works)
3. configure rust-analyzer to use a specific rustfmt package (:RustFmt is borked)

I'm slightly leaning towards 3, CC @NotAShelf for a second opinion

ckoehler:
>     1. revert [fix: disable Conform for Rust #900](https://github.com/NotAShelf/nvf/pull/900), manage rustfmt via conform (:RustFmt is borked)

In my experience it was actually Conform that was broken (not respecting `rustfmt.toml` config); :RustFmt worked fine.

Noah765:
> I'm slightly leaning towards 3, CC @NotAShelf for a second opinion

If there aren't any negative consequences, I think it would make more sense to support both by default. One negative consequence of adding `rustfmt` to `extraPackages` might be confusion over why `:RustFmt` is only supported when `cfg.lsp.enable` or `cfg.dap.enable` is true, i.e. when `rustaceanvim` is installed. However, not supporting `:RustFmt` at all might cause even more confusion because its behavior would be inconsistent across systems, even when everything is enabled. @NotAShelf, what do you think?

> I did have `rustfmt` in my path, but wherever it's coming from, it was actually from the Nix store for me. So I think adding it as an extra package wouldn't add extra space to my install, at least?

Adding `rustfmt` should only increase your installation size if you haven't installed it at all, in which case most users probably want it to be installed, or if you installed a different version than the one `nvf` would install by default. In the latter case, it would be helpful to provide an option to override the package being used.

> So once it's installed, using it should be pretty easy with [config](https://rust-analyzer.github.io/book/configuration#rustfmt.overrideCommand), right?

Once `rustfmt` is installed, formatting should work as expected. The [analyzer settings](https://rust-analyzer.github.io/book/configuration#rustfmt.extraArgs) should also work, because the analyzer delegates formatting to `rustfmt`, and the way that is installed shouldn't matter.

---

On an unrelated note, `clippy` isn't currently installed by default. However, when installed, `rustaceanvim` automatically uses it to provide additional useful lints. I think there should be an `extraDiagnostics` option with `clippy` as a default diagnostics provider. If you agree, I am willing to create a PR to add it. I just need to know if extra diagnostics should always be managed by `nvim-lint`, as is done for other languages. Alternatively, if `clippy` is enabled via `rustaceanvim`, it could potentially eliminate the need to install `nvim-lint`.

ckoehler:
Thanks for doing all this! I don't think I care where rustfmt or clippy come from. I may still use rustup if needed. But as long as I get the latest version of Rust + components, I'll be happy.

ckoehler (approved):

PR #915, opened 2025-05-21
languagaes/neorg: fix grammar packages

Soliprem:
blorb, nixpkgs moment. They deleted neorg grammars.

Currently doesn't build

### DIFF
--- a/docs/release-notes/rl-0.8.md
+++ b/docs/release-notes/rl-0.8.md
@@ -463,3 +463,6 @@
   leverage the pattern introduced in commit [fc8206e7a61d ("flake: utilize
   nix-systems for overridable flake systems")](
   https://github.com/NotAShelf/nvf/commit/fc8206e7a61d7eb02006f9010e62ebdb3336d0d2).
+
+[soliprem](https://github.com/soliprem):
+- fix broken `neorg` grammars
--- a/modules/plugins/notes/neorg/config.nix
+++ b/modules/plugins/notes/neorg/config.nix
@@ -35,7 +35,7 @@ in {
 
     (mkIf cfg.treesitter.enable {
       vim.treesitter.enable = true;
-      vim.treesitter.grammars = [cfg.treesitter.norgPackage];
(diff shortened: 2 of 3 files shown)


NotAShelf (changes requested):
I'm not very keen on accepting a PR that just disables things. It is our responsibility to provide the migration instead of telling the user what to do.

TLDR; figure out why it was removed, package it, and readd the grammar from our own packages.

Soliprem:
the reason it was removed from nixpkgs is that they say that neorg handles the installation on its own, iirc. The reason why I put a disable + warn is that I'm pretty sure this will cause a build failure when we eventually update nixpkgs, and at this time I don't have time to figure out the build because graduation + exam session is coming up irl. So as a temporary solution I thought of disabling + warn. I'll try to see if there's something better that I can get to, though

NotAShelf:
If neorg handles installation on its own, that is bad news for us. We might need to patch the neorg plugin. I'm going to take a look at this myself tomorrow evening.

Soliprem:
was thinking of doing something akin to what nixpkgs did

```nix
norg = pkgs.tree-sitter.buildGrammar {
  language = "norg";
  version = "0.0.0+rev=d89d95a";

  src = pkgs.fetchFromGitHub {
    owner = "nvim-neorg";
    repo = "tree-sitter-norg";
    rev = "d89d95af13d409f30a6c7676387bde311ec4a2c8";
    hash = "sha256-z3h5qMuNKnpQgV62xZ02F5vWEq4VEnm5lxwEnIFu+Rw=";
  };

  meta.homepage = "https://github.com/nvim-neorg/tree-sitter-norg";

};
```
but maybe handling the fetching with npins instead of manually doing it

NotAShelf (approved):
LGTM. I'm assuming you've tested this?

NotAShelf:
Almost forgot, please add an entry to the changelog.

Soliprem:
yep, tested and worked fine on my machine. Would appreciate a second test to make sure it's not some weird "my machine has weirdness" situation though 
(adding changelog now)

NotAShelf (approved):

NotAShelf:
cheers

PR #921, opened 2025-05-27
fix tree-sitter parser packages

BarryLabs:
I guess "nvim-treesitter.builtGrammars" got removed and replaced with "nvim-treesitter-parsers" however it didnt have the norg parser I wanted so I replaced the default options with "pkgs.tree-sitter-grammars" to keep it uniform for now.

### DIFF
--- a/lib/types/languages.nix
+++ b/lib/types/languages.nix
@@ -30,7 +30,7 @@
 
   mkGrammarOption = pkgs: grammar:
     mkPackageOption pkgs ["${grammar} treesitter"] {
-      default = ["vimPlugins" "nvim-treesitter" "builtGrammars" grammar];
+      default = ["tree-sitter-grammars" grammar];
     };
 in {
   inherit diagnostics diagnosticSubmodule mkGrammarOption;
--- a/lib/types/languages.nix
+++ b/lib/types/languages.nix
@@ -30,7 +30,7 @@
 
   mkGrammarOption = pkgs: grammar:
     mkPackageOption pkgs ["${grammar} treesitter"] {
-      default = ["tree-sitter-grammars" grammar];
+      default = ["vimPlugins" "nvim-treesitter" "builtGrammars" grammar];
     };
 in {
   inherit diagnostics diagnosticSubmodule mkGrammarOption;
--- a/configuration.nix
+++ b/configuration.nix
@@ -248,6 +248,7 @@ isMaximal: {
         cmp.enable = isMaximal;
       };
       codecompanion-nvim.enable = false;
+      avante-nvim.enable = isMaximal;
     };
 
     session = {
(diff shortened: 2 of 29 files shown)


NotAShelf (changes requested):
Looks good, thank you for the PR.

THere is one missing function that needs to be handled as well, `mkGrammarOption` found in the extended library will construct package options *explicitly* for `builtGrammars`, which means all language modules will continue trying to pull `builtGrammars`.

BarryLabs:
i think its good now, appreciate the project :)

NotAShelf:
Seems the documentation preview was not updated, could you give me a little bit? I need to double-check something, but I'm off my computer right now. I'll be back in an hour or two, and I can give you a final review.

NotAShelf:
Oops, the CI seems to be failing. If it fails for Nix grammar, I imagine there is an inconsistency in most other language modules as well.

BarryLabs:
!

Ok ill be back

lib/types/languages.nix
```diff
@@ -30,7 +30,7 @@
 
   mkGrammarOption = pkgs: grammar:
     mkPackageOption pkgs ["${grammar} treesitter"] {
-      default = ["vimPlugins" "nvim-treesitter" "builtGrammars" grammar];
+      default = ["tree-sitter-grammars" grammar];
```
horriblename, line 33: as far as I remember pkgs.tree-sitter-grammars does not work as a nvim plugin, are you sure this works?

BarryLabs: I see, I havent had a moment to take a look since.

BarryLabs:
It builds now except for this norg grammar...

We're still referencing pkgs.vimPlugins.nvim-treesitter.norg or whatever it was exactly.

horriblename:
not sure what happened to your git history, but it's including a lot of unrelated changes in [3656f9f](https://github.com/NotAShelf/nvf/pull/921/commits/3656f9f6c5e0c8724f6018313ce703768bf7fd89)

in any case, I'm getting this on trying to run `.#minimal`:

```
error: A definition for option `vim.startPlugins."[definition 36-entry 1]"' is not of type `null or package or one of "blink-cmp", "aerial-nvim", "alpha-nvim", ...(shortened)..., "vim-repeat"'. Definition values:
- In `/nix/store/m89sr4423yaxby8jqz6363lbai7qgcbb-source/modules/plugins/treesitter/config.nix': "nvim-treesitter"
```

BarryLabs:
I went crazy with the builder functions and reversed most of them such that it builds my current config but CSS "prettier" and the norgpackage functions are still issues.

I was testing it live and I settled here as I didnt want to alter a large portion of the codebase without at least some conversation about it.

I am quite stumped how when I hardcode the norg grammar as a default in the "change default?" change, it still expects the previous package...?

PR #935, opened 2025-06-03
build: fix plugin overrides from fetchers

horriblename:
This is actually the worst PR I have ever made in my life


horriblename:
superseded by 939

PR #937, opened 2025-06-04
fix(blink-cmp): Extended auto_show type

itpropro:
See https://cmp.saghen.dev/modes/cmdline.html#show-menu-automatically on how to add a function for more control of when the menu actually opens.

### DIFF
--- a/docs/release-notes/rl-0.8.md
+++ b/docs/release-notes/rl-0.8.md
@@ -423,3 +423,7 @@
 
 - Add missing `transparent` option for existing
   [onedark.nvim](https://github.com/navarasu/onedark.nvim) theme.
+  
+[theutz](https://github.com/theutz):
+
+- Added "auto" flavour for catppuccin theme
--- a/modules/plugins/completion/blink-cmp/blink-cmp.nix
+++ b/modules/plugins/completion/blink-cmp/blink-cmp.nix
@@ -10,7 +10,7 @@
     freeformType = attrsOf (listOf (either str luaInline));
     options = {
       preset = mkOption {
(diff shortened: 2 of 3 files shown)


itpropro:
This should also apply to `cmdline.completion.menu.auto_show`, but the cmdline section is not really implemented yet. Should I create a separate PR for that @NotAShelf ?

horriblename:
> This should also apply to cmdline.completion.menu.auto_show, but the cmdline section is not really implemented yet. Should I create a separate PR for that

you can but it's not required. The point of setupOpts is so we don't have to add every plugin option out there

horriblename (approved):

NotAShelf:
Please ensure that the failing formatting checks (editorconfig and alejandra) pass before we can merge this.

itpropro:
> > This should also apply to cmdline.completion.menu.auto_show, but the cmdline section is not really implemented yet. Should I create a separate PR for that
> 
> you can but it's not required. The point of setupOpts is so we don't have to add every plugin option out there

The problem is that this is not consistent. SetupOpts in the blink case cannot be everything, as specific things under there are overwritten by custom options like the presets. So at least the options that are implemented should allow a high amount of customization or a way to overwrite them if need to be adjusted.

horriblename:
> The problem is that this is not consistent. SetupOpts in the blink case cannot be everything, as specific things under there are overwritten by custom options like the presets. So at least the options that are implemented should allow a high amount of customization or a way to overwrite them if need to be adjusted.

Sorry, I don't follow, can you give an example?

Issue #929, opened 2025-05-31
Overriding codecompanion plugin package fails


puzzleshark:
### ⚠️ Please verify that this bug has NOT been reported before.

- [x] I checked all existing issues and didn't find a similar issue

### Description

This configuration fails for me:
```
      packages.x86_64-linux.broken = (inputs.nvf.lib.neovimConfiguration {
        pkgs = inputs.nixpkgs.legacyPackages.x86_64-linux;
        modules = [
          {
            vim.pluginOverrides = {
              codecompanion-nvim = inputs.nixpkgs.legacyPackages.x86_64-linux.fetchFromGitHub {
                owner = "olimorris";
                repo = "codecompanion.nvim";
                rev = "29ded250b186c550222aa6a4a648e52ad49dd85f";
                sha256 = "0l98qgdn4lk59ri2qvdiq1y3yzrdwsyswvmrswb5xx8j2mmzci35";
              };
            };
            vim.assistant.codecompanion-nvim.enable = true;
          }
        ];
      }).neovim;
```

with:
```❯ nix run .#broken
error:
       … while calling the 'derivationStrict' builtin
         at <nix/derivation-internal.nix>:37:12:
           36|
           37|   strict = derivationStrict drvAttrs;
             |            ^
           38|

       … while evaluating derivation 'nvf-with-helpers'
         whose name attribute is located at /nix/store/2q205l7a46885nq31vvji40m30bygzzj-source/pkgs/stdenv/generic/make-derivation.nix:480:13

       … while evaluating attribute 'paths' of derivation 'nvf-with-helpers'
         at /nix/store/2q205l7a46885nq31vvji40m30bygzzj-source/pkgs/build-support/trivial-builders/default.nix:619:11:
          618|           inherit preferLocalBuild allowSubstitutes;
          619|           paths = mapPaths (path: "${path}${stripPrefix}") paths;
             |           ^
          620|           passAsFile = [ "paths" ];

       … while evaluating the option `plugins.opt':

       … while evaluating definitions from `<unknown-file>':

       (stack trace truncated; use '--show-trace' to show the full, detailed trace)

       error: expected a set but found a string with context: "/nix/store/wmzjhg9b8pnm2g42yrpk34m5054way53-source"


### 👟 Reproduction steps

Add the above configuration in a flake.nix and try to build/run

### 👀 Expected behavior

I would expect the above configuration to work

### 😓 Actual Behavior

the configuration fails

### 💻 Metadata

[...]

PR #939, opened 2025-06-04
fix: pluginOverrides attempting to build

Gerg-L:
closes: #929
supersedes: #935

### DIFF
--- a/modules/wrapper/build/config.nix
+++ b/modules/wrapper/build/config.nix
@@ -7,9 +7,9 @@
 }: let
   inherit (pkgs) vimPlugins;
   inherit (lib.trivial) flip;
-  inherit (builtins) filter isString;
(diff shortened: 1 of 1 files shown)


modules/wrapper/build/config.nix
```diff
@@ -53,7 +53,9 @@
 
   buildConfigPlugins = plugins:
     map (plug:
-      if (isString plug)
+      if hasAttr plug config.vim.pluginOverrides
```
horriblename:
```suggestion
      if isString plug && hasAttr plug config.vim.pluginOverrides
```

horriblename (changes requested):

horriblename (approved):
non-flake inputs still broken but will fix another time

PR #934, opened 2025-06-03
fix, Add dependencies needed for building Avante

Munsio:
Fixes #933

PR #941, opened 2025-06-06
avante-nvim: remove unnecessary dependencies in plugin build

Munsio:
As described in https://github.com/NotAShelf/nvf/pull/934#issuecomment-2937368328 this is my take to remove the unwanted dependencies during the vim plugin build process.

My guess is that the plugin build is only necessary because of some shared libraries (flake/avante-nvim/default.nix:52), unfortunately I was unable to find another way to set them for the plugin itself.

If anyone can show some insights on how a change would look like I am happy to try it again, otherwise this is the best I can do for now.

### DIFF
--- a/flake/avante-nvim/default.nix
+++ b/flake/avante-nvim/default.nix
@@ -1,16 +1,14 @@
 {
-  nix-update-script,
   openssl,
   pkg-config,
   rustPlatform,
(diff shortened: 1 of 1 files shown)


NotAShelf:
Thank you :)

PR #960, opened 2025-06-16
avante-nvim: Migrate provider options

Munsio:
Migrate the provider options according to the migration guide. https://github.com/yetone/avante.nvim/wiki/Provider-configuration-migration-guide


horriblename (approved):
LGTM, thanks!

horriblename:
I assume the current version of our avante uses the new provider API? (and that we don't need to update avante for this to work)

Munsio:
> I assume the current version of our avante uses the new provider API? (and that we don't need to update avante for this to work)

Yes it got updated with eacce4eb022171211baaad7ba7fc6cf561773f0e 

I found out about it yesterday after my os update chore.

PR #962, opened 2025-06-17
[skip ci] Language overhaul part 2

horriblename:
part of #913 

CI skipped because it's trying to build docs from an old commit

### DIFF
--- a/modules/extra/deprecations.nix
+++ b/modules/extra/deprecations.nix
@@ -1,5 +1,6 @@
 {lib, ...}: let
-  inherit (lib.modules) mkRemovedOptionModule mkRenamedOptionModule;
+  inherit (builtins) head warn;
+  inherit (lib.modules) mkRemovedOptionModule mkRenamedOptionModule doRename;
   inherit (lib.lists) concatLists;
   inherit (lib.nvim.config) batchRenameOptions;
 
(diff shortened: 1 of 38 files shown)


modules/plugins/languages/asm.nix
```diff
@@ -34,7 +34,7 @@ in {
     lsp = {
       enable = mkEnableOption "Assembly LSP support" // {default = config.vim.lsp.enable;};
       servers = mkOption {
-        type = listOf (enum (attrNames servers));
+        type = singleOrListOf (enum (attrNames servers));
         default = defaultServers;
         description = "Assembly LSP server to use";
       };
```
imnotpoz, line 40: these are the same across all languages and can be extracted into a custom `mkServersOption`, as taken from my closed PR:

```nix
mkServersOption = language: servers: defaultServers:
  mkOption {
    type = singleOrListOf (enum (attrNames servers));
    default = defaultServers;
    description = "${language} LSP servers to use";
  };
```

under `lib.nvim.types` with the following usage

```nix
lsp = {
  enable = mkEnableOption "Assembly LSP support" // {default = config.vim.lsp.enable;};
  servers = mkServersOption "Assembly" servers defaultServers;
};

Comment (horriblename) [reply modules/plugins/languages/asm.nix:40]: only some are changed to singleOrListOf where needed, (not all language modules have the old `server` option). I think I'll add a warning for using it as a single string, unless @NotAShelf wants all of them to be singleOrListOf

**Comment by alfarelcynthesis** [comment]
I was just testing this out and noticed that the `tailwindcss` module still uses `lspconfig.util` for `root_dir`, even though `lspconfig` isn't necessarily enabled.

**Comment by horriblename** [comment]
nice catch, thanks!

File: modules/extra/deprecations.nix [hunk 2]
```diff
@@ -20,6 +21,27 @@
...
+  mkRenamedLspServer = lang:
+    doRename
+    {
+      from = ["vim" "languages" lang "lsp" "server"];
+      to = ["vim" "languages" lang "lsp" "servers"];
+      visible = false;
+      warn = true;
+      use = x:
+        warn
+        "Obsolete option `vim.languages.${lang}.lsp.server` used, use `${lang}.lsp.servers` instead."
```
Soliprem: I assume this is meant to be vim.languages.${lang}.lsp.servers (imo we should be explicit about it to avoid confusion)

horriblename: yep, good catch

PR #972, opened 2025-06-26
binds/which-key: allow deregistering binds or groups

lackac:
`options.vim.binds.whichKey.register` has type `attrsOf (nullOr str)`, but it actually failed to build with `null` values. This fixes that and uses null values as a signal to not register a bind or group that was set by one of the plugins.

## Sanity Checking

- Style and consistency
- Package(s) built:
  - [x] `.#maximal`
- Tested on platform(s)
  - [ ] `x86_64-linux`
  - [ ] `aarch64-linux`
  - [ ] `x86_64-darwin`
  - [x] `aarch64-darwin`

---


horriblename:
this doesn't "deregister" the group does it? just makes it a group with no description

lackac:
> this doesn't "deregister" the group does it? just makes it a group with no description

no, it does. It uses `lib.lists.optional` to omit those keys that are set to `null` from the actual configuration that which-key receives. There are a number of keys that are registered in the nvf config (e.g. [for trouble](https://github.com/NotAShelf/nvf/blob/5bad5dd94ce5ea3b40b08d9e6802e69d02198d21/modules/plugins/lsp/trouble/config.nix#L32-L35)) and they use `pushDownDefault` so can be overridden by setting them `null` by the user. The change in this PR is necessary to make that possible.

modules/plugins/utility/binds/which-key/config.nix
```diff
@@ -10,7 +10,7 @@
   inherit (lib.nvim.dag) entryAnywhere;
 
   cfg = config.vim.binds.whichKey;
-  register = mapAttrsToList (n: v: mkLuaInline "{ '${n}', desc = '${v}' }") cfg.register;
+  register = mapAttrsToList (n: v: lib.lists.optional (! isNull v) (mkLuaInline "{ '${n}', desc = '${v}' }")) cfg.register;
```
horriblename, line 13: lib.lists.optional returns a list, so we'll get something like

```lua
wk.add({
  { { 'a', desc = 'desc' } },
  { { 'b', desc = 'desc' } },
})
```

just want to confirm this works as intended

lackac: Good question, but it looks like this works. I tested it another time and added two more registrations to my own config. The resulting lua code is

```lua
wk.add({
  {{ '<leader>g', desc = '+Git' }},
  {{ '<leader>h', desc = '+Gitsigns' }},
  {{ '<leader>k', desc = '+KKKKkkkk' }},
  {},
  {},
  {{ '<leader>z', desc = '+Zzz' }}
})
```

The empty tables are likely the filtered out registrations which I set to `null`. I could see all groups that were left in the which-key UI.

PR #975, opened 2025-06-27
Update vim.clipboard.registers type to str

vaclavek-kamil:
nvf.settings.vim.clipboard.registers allows for either a string or a list of strings, but Neovim only accepts a string value. When a NVF user follows the options Appendix and passes a string, this results in an invalid Neovim config. Solved by updating the type to allow strings only.

### DIFF
--- a/modules/neovim/init/clipboard.nix
+++ b/modules/neovim/init/clipboard.nix
@@ -20,7 +20,7 @@ in {
         '';
 
         registers = mkOption {
-          type = either str (listOf str);
+          type = str;
           default = "";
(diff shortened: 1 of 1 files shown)


vac-kam:
Btw I'd assume that the options Appendix is dynamically rendered from the code, am I right?

rrvsh (approved):

rrvsh:
Tested and works for me

NotAShelf (approved):

NotAShelf:
Can't merge on my phone for some reason, @horriblename could you do the honors?

PR #984, opened 2025-07-01
Added logic to configure nil_ls via the same options parameter used for nixd.

ItsLiyua:
I got fed up with being asked whether I want to "fetch inputs" every time I open a nix file so I added the logic to configure nil_ls via `languages.nix.lsp.options` the same way that nixd is configured. I'm somewhat new to nix so if the code doesn't have the quality you'd like it to have tell me so I can try fixing it myself. You might see some more changes than needed because I have my nvf set up in a way to auto format files on save and I use nixfmt so it might have gotten messed up in some of the commits. I made sure to check that it's formatted with alejandra in the last commit tho.

I hope this helps! Thanks for reading!

### DIFF
--- a/modules/plugins/languages/nix.nix
+++ b/modules/plugins/languages/nix.nix
@@ -4,8 +4,7 @@
   lib,
   inputs,
   ...
-}:
-let
+}: let
   inherit (builtins) attrNames;
   inherit (lib) concatStringsSep;
   inherit (lib.meta) getExe;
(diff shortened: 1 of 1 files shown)


horriblename:
we're close to finishing languages overhaul in #913, after which you can do this yourself via `vim.lsp.servers.nil.settings...`, and `cfg.lsp.options` in the nix module will be removed. So I don't want to merge this.

for the time being I _think_ it would work if you put your settings directly in `vim.lsp.servers.nil.settings...`. I'm not really sure because I don't know the specifics of how it interacts with lspconfig

ItsLiyua:
Alright. Thanks for telling me about an alternative fix :D

PR #990, opened 2025-07-05
assistant/supermaven-nvim: init

justDeeevin:
supermaven-nvim's README claims nvim-cmp integration is as simple as adding `"supermaven"` to the sources list, so I felt it unnecessary to add a separate option for that here. Let me know if that should be changed.

Also, this is my first time contributing here, so sorry if there's anything that doesn't fit standards. I referred to the other assistant options when writing this.

### DIFF
--- a/docs/manual/configuring.md
+++ b/docs/manual/configuring.md
@@ -8,7 +8,7 @@ your convenience. You might also be interested in the [helpful tips section] for
 more advanced or unusual configuration options supported by nvf.
 
 Note that this section does not cover module _options_. For an overview of all
-module options provided by nvf, please visit the [appendix](/options.html)
+module options provided by nvf, please visit the [appendix](/nvf/options.html)
 
 ```{=include=} chapters
 configuring/custom-package.md
--- a/docs/release-notes/rl-0.8.md
+++ b/docs/release-notes/rl-0.8.md
@@ -454,3 +454,12 @@
 - Add [supermaven-nvim] plugin in `vim.assistant.supermaven-nvim` with `enable` and
   `setupOpts`
 
+[trueNAHO](https://github.com/trueNAHO):
+
+- `flake-parts`'s `nixpkgs-lib` input follows nvf's `nixpkgs` input to reduce
+  download size.
+
+- `flake-utils`'s `systems` inputs follows nvf's `systems` input to transitively
+  leverage the pattern introduced in commit [fc8206e7a61d ("flake: utilize
+  nix-systems for overridable flake systems")](
+  https://github.com/NotAShelf/nvf/commit/fc8206e7a61d7eb02006f9010e62ebdb3336d0d2).
--- a/flake.lock
+++ b/flake.lock
@@ -2,7 +2,9 @@
   "nodes": {
     "flake-parts": {
       "inputs": {
-        "nixpkgs-lib": "nixpkgs-lib"
+        "nixpkgs-lib": [
+          "nixpkgs"
+        ]
       },
       "locked": {
         "lastModified": 1749398372,
@@ -20,7 +22,9 @@
     },
     "flake-utils": {
       "inputs": {
-        "systems": "systems"
+        "systems": [
+          "systems"
+        ]
       },
       "locked": {
         "lastModified": 1731533236,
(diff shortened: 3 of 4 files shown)


modules/plugins/assistant/supermaven/config.nix
```diff
@@ -0,0 +1,17 @@
+{
+  config,
+  lib,
+  ...
+}: let
+  cfg = config.vim.assistant.supermaven;
+in {
+  config = lib.mkIf cfg.enable {
+    vim.plugins = {
```
Soliprem: vim.plugins needs to be `vim.lazy.plugins`

modules/plugins/assistant/supermaven/supermaven.nix
```diff
@@ -0,0 +1,82 @@
+{lib, ...}: let
+  inherit (lib) mkOption mkEnableOption types;
+in {
+  options.vim.assistant.supermaven = {
```
Soliprem: should be `supermaven-nvim` if that's the plugin name

modules/plugins/assistant/supermaven/config.nix
```diff
@@ -0,0 +1,17 @@
+{
+  config,
+  lib,
+  ...
+}: let
+  cfg = config.vim.assistant.supermaven;
```
Soliprem: supermaven-nvim

Soliprem (changes requested):
off the bat, that change will get it to build. On the rest, I'm unfamiliar with the workflow for Supermaven, but on the nix side (besides my reviews), things LGTM.

I recommend you use the checklist, regarding style and testing, that comes as the templates for new PRs :). It's great for first-time contributors to dodge "obvious" reviews and for long-time contributors as an actual checklist of things to remember (it's also just a bunch of things we don't have to wonder if you did, which makes reviewing easier).

Thanks for the work <3

p.s. It looks like it works correctly, but I noticed it downloading a binary? Unsure of where that's being stored

modules/plugins/assistant/supermaven/supermaven.nix
```diff
@@ -0,0 +1,82 @@
+{lib, ...}: let
+  inherit (lib) mkOption mkEnableOption types;
```
NotAShelf:
```suggestion
  inherit (lib.options) mkOption mkEnableOption
  inherit (lib.types) ...;
```

Please be explicit with the imports.

modules/plugins/assistant/supermaven/supermaven.nix
```diff
@@ -0,0 +1,82 @@
...
+      };
+      disable_keymaps = mkOption {
+        type = types.nullOr types.bool;
+        default = null;
+        description = "Disable built-in keymaps for more manual control";
+      };
+      condition = mkOption {
+        type = types.nullOr lib.nvim.types.luaInline;
+        default = null;
+        description = "Condition function to check for stopping supermaven. A returned `true` means to stop supermaven";
```
NotAShelf:
```suggestion
        description = ''
          Condition function to check for stopping supermaven.
          A returned `true` means to stop supermaven
        '';
```

Multi-lining long strings is better for readibility.

NotAShelf (changes requested):
Thank you for the good work. I'm leaving two comments for the sake of consistency with the project styles. Soliprem's comments also apply.

Lastly, please add a changelog entry before we can merge this.

justDeeevin:
> I recommend you use the checklist

When I created the PR it didn't pre-fill the description with the template. Don't know why that happened! Thanks for your feedback in the face of this.

Also, apologies for not properly testing, I should have known better. I need to stop making midnight PRs🫠

modules/plugins/assistant/supermaven/supermaven.nix
justDeeevin: Didn't previously know about `lib.options`.

Oh nixpkgs lib how I love/hate you...

justDeeevin: Would you like me to import each type I use or can I keep using the `types` attrset directly?

justDeeevin:
I changed it to `supermaven-nvim` as requested. However, I still think `supermaven` would make more sense. The option for copilot is just `assistant.copilot` even though the plugin itself is called `copilot-lua`. Is there a particular rationale for this I'm not seeing?

modules/plugins/assistant/supermaven/supermaven.nix
NotAShelf: I'd prefer if you imported each type for future extensibility without annoying diffs or `with types;` like hacks.

justDeeevin:
> p.s. It looks like it works correctly, but I noticed it downloading a binary? Unsure of where that's being stored

Reading the source code, it looks like this gets put in `$XDG_DATA_HOME/supermaven` (or `$HOME/.supermaven` if `XDG_DATA_HOME` is unset).

NotAShelf:
> I changed it to supermaven-nvim as requested. However, I still think supermaven would make more sense. The option for copilot is just assistant.copilot even though the plugin itself is called copilot-lua. Is there a particular rationale for this I'm not seeing?

Sorry for the late response. This is a relatively new "convention" that we've started applying in light of some future refactor plans. There are plugins that do *not* adhere to the convention yet, as they have been added *before* we even discussed such a naming decision. It'll be easier for everyone if you name the options after the plugins :)

lackac:
> When I created the PR it didn't pre-fill the description with the template. Don't know why that happened! Thanks for your feedback in the face of this.

I also noticed this. I'm always adding `&template=pull_request_template.md` to the end of the URL when creating a pull request. But if we want people to use the checklist, it should be automatic. I think the file could be moved to `.github/pull_request_template.md` if a single template is sufficient and that will make it apply automatically.

NotAShelf:
I'm quite positive that the pull request template is active by default. Unless, of course, GH has changed how it functions. Last I've checked the template is "prompted" by default, and it's usable in CLI tools like `gh`. I'll look into it on monday.

justDeeevin:
What's left to do here?

NotAShelf (approved):

NotAShelf:
can be merged with @Soliprem's approval.

Soliprem (dismissed):
LGTM

justDeeevin:
CI was failing due to an oversight when renaming from supermaven to supermaven-nvim. fix pushed. Also force-pushed with rebase to resolve conflict.

NotAShelf:
I'm sorry I've caused a merge conflict again, if you could resolve it I can merge this right away

justDeeevin:
So strange that that should cause a conflict.

Done!

NotAShelf:
Appears that the CI is failing, could you please change the mkPluginSetupOption line to a `inherit (lib.nvim.types) mkPluginSetupOption;` at the top and then consume just the function in config?

NotAShelf:
Looks like the CI is still failing, could you run `nix build .#docs-html` to confirm that all CI steps are passing after fixing the current failure? We should be able to merge then.

justDeeevin:
:sob:

justDeeevin:
Would you prefer I merge instead of rebasing to resolve conflicts?

NotAShelf:
A quick resolution should be fine, there aren't any more incoming PRs atm :)

justDeeevin:
there

NotAShelf (approved):

NotAShelf:
Awesome work, thank you :)

justDeeevin:
Sorry it's taken so long :sob:

PR #1030, opened 2025-07-20
flake: move packages to a by-name overlay

NotAShelf:
Moves plugin packages to a nested packages directory with a `by-name` overlay for organization and scoping. This makes packaging new plugins slightly more convenient, and the layout more intuitive. I've also documented the addition of new plugins, with other minor tweaks to the npins section.

Lastly, the wording for `toLuaObject` has been improved and some examples have been added. I'll need to approval of both @horriblename and @Soliprem for this.


docs/manual/hacking/additional-plugins.md
```diff
@@ -1,25 +1,92 @@
...
+to fetch the plugin sources somehow. npins is, once again, the recommended
+option to fetch the plugin sources. Refer to the previous section on how to use
+npins to add a new plugin.
 
-You can now reference this plugin as a **string**.
+Plugins built from source must go into the `flake/pkgs/by-name` overlay. It will
+automatically create flake outputs for individual packages. Lastly, you must add
+your package to the plugin builder (`pluginBuilders`) function manually in
+`modules/wrapper/build/config.nix`. Once done, you may refer to your plugin as a
+**string**.
```
horriblename, line 84: should you mention that it needs the `pkgs/by-name/...` to be `git add`-ed? or is it a common enough thing that nix users know

NotAShelf: That applies to all new additions, so I will hope that the average contributor knows how flakes operate...

horriblename: I was concerned cuz idk what the error would look like, if it was "`path/to/file.nix` doesn't exist" we're already conditioned to `git add`, but I'm guessing the error would be "attribute `plugin-name` does not exist", which is probably less... intuitive? I'll let you decide if you wanna add it or not I'm probably overthinking this lol

NotAShelf: I'm in favor of crossing that bridge when we get to it, lol

PR #1051, opened 2025-07-27
Fix yaml-language-server on_attach

alfarelcynthesis:
The current (v0.8) configuration errors when the yaml language module is enabled.

## Sanity Checking

- Style and consistency
- Package(s) built:
  - [ ] `.#maximal`
    - Would take a while and yaml isn't enabled here.
- Tested on platform(s)
  - [x] `x86_64-linux`
    - I tested that it parses correctly into the function, but not on an actual yaml file.
  - [ ] `aarch64-linux`
  - [ ] `x86_64-darwin`
  - [ ] `aarch64-darwin`

Note: formatting seems broken via `nix fmt` (hangs waiting for stdin input?)

---


NotAShelf (commented):
I'd actually prefer if the onAttach variable was kept for brevity. It'll be a smaller diff, and it seems the problem is the lack of `mkLuaInline`; no need to move.

For the formatter, it has to be `nix fmt .` as of recent Nix versions. I made an incorrect assumption that all users are familiar with Nix's quirks, I'll push a fix to update the formatter.

alfarelcynthesis:
> I'd actually prefer if the onAttach variable was kept for brevity. It'll be a smaller diff, and it seems the problem is the lack of `mkLuaInline`; no need to move.
> 
> For the formatter, it has to be `nix fmt .` as of recent Nix versions. I made an incorrect assumption that all users are familiar with Nix's quirks, I'll push a fix to update the formatter.

Thanks for the formatter tip, I honestly assumed supplying `.` would be to select the flake, which I thought should be autodetected correctly.

Is it okay to clean up the declaration a little bit using an inherit, or should I leave it exactly the same other than the fix?

___

Comments after force pushing:

- When I was retesting things I noticed that when `vim.lsp.enable` was true, the `helm.lsp` check triggered even when `helm.enable` wasn't true. It doesn't break anything, but I still fixed it.

- I tested this with helm enabled and found another bug, which I fixed in the second commit.

- The `yaml-language-server` helm `on_attach` function doesn't call the default one, so I'm not sure if it correctly registers keymaps, given I don't really know what that would do in this case or how to test it.

alfarelcynthesis:
Closing and reopening after rebase, this doesn't seem to have updated automatically for some reason and is failing because of an old docs issue elsewhere.

PR #1094, opened 2025-08-21
docs: extend configuring section with details on the module system

### LINKED ISSUE #1067: utility/oil-nvim: key-value pair
Hey,

I am having some trouble creating the key-values pairs for the setupOpts. I dont really understaind how the conversion works.
Mostly with the keymaps.
In the lua code, there is no key for the first element, but I guess the setupOpts requires one. So I dont really know what to do.

Nix code:
```
  setupOpts = {
      columns = [ "icon" ];
      keymaps = {
        "g?" = { "1" = "actions.show_help"; mode = "n"; };
        "<CR>" = "actions.select";
        "<C-p>" = "actions.preview";
        "<C-c>" = { "1" = "actions.close"; mode = "n"; };
        "<C-l>" = "actions.refresh";
        "<BS>" = { "1" = "actions.parent"; mode = "n"; };
        "<A-BS>" = { "1" = "actions.open_cwd"; mode = "n"; };
        "g." = { "1" = "actions.toggle_hidden"; mode = "n"; };
      };
      use_default_keymaps = false;
      view_options = {
        show_hidden = true;
      };
    };
```

original lua code:
```
		require("oil").setup({
			columns = {
				"icon",
				-- "permissions",
				-- "size",
				-- "mtime",
			},
			keymaps = {
				["g?"] = { "actions.show_help", mode = "n" },
				["<CR>"] = "actions.select",
				["<C-p>"] = "actions.preview",
				["<C-c>"] = { "actions.close", mode = "n" },
				["<C-l>"] = "actions.refresh",
				["<BS>"] = { "actions.parent", mode = "n" },
				["<A-BS>"] = { "actions.open_cwd", mode = "n" },
				["g."] = { "actions.toggle_hidden", mode = "n" },
				["gd"] = {
					desc = "Toggle file detail view",
					callback = function()
						detail = not detail
						if detail then
							require("oil").set_columns({ "icon", "permissions", "size", "mtime" })
						else
							require("oil").set_columns({ "icon" })
						end
					end,
				},
			},
			use_default_keymaps = false,
			view_options = {
				show_hidden = true,
			},
		})
```

NotAShelf:
This is an initial draft of a new `modules` section describing how we use the Nix module system within nvf. While it **does not** cover every module, I think it is worth covering the *basics* of a module to give the users and idea and to potentially answer questions from those New to the Nix ecosystem.

I also think we stand to gain from adding some examples in the DAG section, but I reckon it may be more confusing for the users than it is helpful. CC @horriblename and @Soliprem for reviews and feedback. This *can* be merged as is, but worth discussing beforehand.

Closes #1067

### DIFF
--- a/docs/manual/configuring/lazy-loading/overview.md
+++ b/docs/manual/configuring/lazy-loading/overview.md
@@ -1,0 +1,51 @@
+# Overview {#sec-lazy-overview}
+
+Lazy loading defers plugin initialization until the plugin is actually needed.
+Rather than loading every plugin at startup, nvf (via its `lz.n` backend) loads
+plugins on demand, triggered by commands, events, filetypes, or keymaps. This
+reduces startup time, especially in configurations with many plugins.
+
+## Why Should I LazyLoad? {#sec-why-get-lazy}
+
+Neovim evaluates all sourced Lua and Vimscript at startup. Plugins that register
+autocommands, define highlight groups, or call `require()` unconditionally all
+contribute to startup latency. For frequently opened editors (terminals,
+`git commit`, quick file edits) even a few hundred milliseconds is noticeable.
+Lazy loading moves that cost to the first use of a feature instead of every
+startup.
+
+## lz.n {#sec-lazy-backend}
+
+[lz.n]: https://github.com/lumen-oss/lz.n
+
+nvf uses [lz.n] as its lazy-loading backend. `lz.n` is a minimal, declarative
+lazy loader for Neovim that integrates with Nix-managed plugin paths. Unlike
+runtime package managers, `lz.n` does not download or manage plugins; it only
+controls when they are sourced. This plays nicely with our model of using Nix to
+manage plugins.
+
+## Enabling Lazy Loading {#sec-lazy-enable}
+
+Lazy loading is configured via `vim.lazy.plugins`, which accepts an attribute
+set mapping plugin names to lazy-loading specifications:
+
+```nix
+{
+  config.vim.lazy.plugins = {
+    "my-plugin.nvim" = {
+      package = pkgs.vimPlugins.my-plugin-nvim;
+
+      # Mark the plugin as lazy. It will not load at startup.
+      lazy = true;
+
+      # Define at least one trigger to load the plugin.
+      cmd = [ "MyPluginCommand" ];
+    };
+  };
+}
+```
+
+If no trigger (`cmd`, `event`, `keys`, `ft`) is specified alongside
+`lazy =
+true`, the plugin will never be automatically loaded. You must call
+`require("lz.n").load("my-plugin.nvim")` manually or define a trigger.
--- a/docs/manual/configuring/modules.md
+++ b/docs/manual/configuring/modules.md
@@ -1,0 +1,58 @@
+## Using the Module Interface {#ch-module-interface}
+
+Before describing the module interface, it is worth noting that NVF is a hybrid
+wrapper. It does not lock you into one of Lua or Nix, and both languages are
+considered first-class citizens for configuring your editor. However, Nix is the
+primarily supported language for NVF. While [DAGs](#ch-using-dags) allow for the
+surgical insertion of Lua code into your configuration, in most cases you will
+be more interested in using or extending the Nix-based module system.
+
+### {#ch-using-modules}
+
+Up until v0.6, most modules exposed all supported plugin options as individual
+module options. With the release of v0.6, almost every module has been converted
+to a new `setupOpts` format that provides complete user freedom over a plugin's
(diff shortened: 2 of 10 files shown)


docs/manual/configuring/modules.md
```diff
@@ -0,0 +1,58 @@
+## Using the Module Interface {#ch-module-interface}
```
horriblename, line 1: I feel like this page should be named "configuring supported plugins" or at least something that mentions plugins

docs/manual/hacking/additional-plugins.md
```diff
@@ -165,15 +165,33 @@ own fields!
...
    - `{foo = "bar"}` -> `{["foo"] = "bar"}`
    - `["foo" "bar"]` -> `{"foo", "bar"}`
+   - You may also write **mixed tables** using `toLuaObject`, using a special
+     syntax to describe a key's position in the table. Let's say you want to get
+     something like `{"foo", bar = "baz"}` expressed in Lua using Nix. The
+     appropriate Nix syntax to express mixed tables is as follows:
+
+     ```nix
+     # Notice the position indicator, "@1"
+     { "@1" = "foo"; bar = "baz"; };
```
horriblename: uuuhhh the "`@1`" syntax was never really exposed, and has been an internal thing thus far. I'd really like to address some problems with the current design before we actually expose them, namely:

1. you can't "escape" the @ syntax, i.e. you can't make a lua table like this: `{ ["@1"] = something }`
2. being able to add raw lua expressions as values - this is purely so we can express the lua expr `{ [vim.log.levels.ERROR] = "E" }` in nix
3. I also don't like the random ass `@` syntax - I'd rather use something like this: (this is in nix) `{ "[1]" = something; }`. At least this looks like the lua syntax

to that end, I'd like to propose we drop the `@` syntax, and use something like this:

```
# this nix attrset:
{
  "[1]" = 1;
  "[vim.log.levels.ERROR]" = "E";
  ''["[example to show how to escape brackets]"]'' = 2;
}
# converts to this lua table:
{
  [1] = 1;
  [vim.log.levels.ERROR] = "E";
  ["[example to show how to escape brackets]"] = 2;
}
```

I'm still a bit iffy on the `[...]` syntax, but even if we stick to `@` I'd rather interpret it as a raw lua expression since it solves 1. and 2.

regardless, both options are breaking changes I should probably put in a separate issue

all this to say, please remove this section until we fix the issues with the current syntax :)

horriblename (changes requested):

snoweuph:
I was so free to make this up to date with main

snoweuph:
also I was so free to run `nix fmt` on this

docs/manual/configuring/modules.md
NotAShelf: I think we should not make this page exclusive to plugins. It's about the modules, and modules can be about more than just plugins.

1. You can write your *own* modules and vendor them for others
2. You can use `vim.*` to configure vanilla Neovim behaviour
3. nvf is mostly about modules for everything, not just plugins

I think "configuring supported plugins" heavily narrows the scope of this page for no good reason.

docs/manual/configuring/lazy-loading/overview.md
```diff
@@ -0,0 +1,51 @@
+# Overview {#sec-lazy-overview}
+
+Lazy loading defers plugin initialization until the plugin is actually needed.
+Rather than loading every plugin at startup, nvf (via its `lz.n` backend) loads
+plugins on demand, triggered by commands, events, filetypes, or keymaps. This
+reduces startup time, especially in configurations with many plugins.
+
+## Why Should I LazyLoad? {#sec-why-get-lazy}
```
horriblename, line 8: > LazyLoad

is that a typo or intentional?

docs/manual/configuring/lazy-loading/overview.md
```diff
@@ -0,0 +1,51 @@
...
+
+      # Define at least one trigger to load the plugin.
+      cmd = [ "MyPluginCommand" ];
+    };
+  };
+}
+```
+
+If no trigger (`cmd`, `event`, `keys`, `ft`) is specified alongside
+`lazy =
```
horriblename, line 49: extra newline that (probably?) messes up the HTML formatting

horriblename (changes requested):

docs/manual/configuring/lazy-loading/overview.md
NotAShelf: It's intentional. We can change, though. I've always seen it get referred to as "LazyLoading" in docs since it's like a trademark term of sorts.

horriblename (approved):

PR #1095, opened 2025-08-21
 languages/rust: move crates-nvim dependency to extensions attribute [v0.8]

NotAShelf:
Moves `languages.rust.crates` from top-level to `languages.rust.extensions.crates-nvim` to match the recently introduced "extensions" idiom, and switches to `setupOpts` for configuring crates.nvim. This also "fixes" a bug with completion sources by enabling completion sources if nvim-cmp is enabled. Worth noting that the nvim-cmp integration has been removed in the master branch of crates.nvim, so the workaround can be removed by opting into crates.nvim's built-in LSP option once we update the plugin. An alternative to the "what if we forget" is to update it to master now, but I'd rather not.


modules/plugins/languages/rust.nix
```diff
@@ -103,25 +93,32 @@ in {
...
+          description = ''
+            Whether to add crates.nvim as a source for completion plugins. The following
+            plugins are supported by crates.nvim:
+
+            * nvim-cmp
+            * coq.nvim
+
+            However nvf only supports auto-setup for nvim-cmp.
+          '';
+        };
```
horriblename: why not add it under `setupOpts.cmp.enabled`?

horriblename (changes requested):

modules/plugins/languages/rust.nix
NotAShelf: I forgor 💀 

much better idea though, fixed

PR #1101, opened 2025-08-22
lsp/harper-ls: init

gmvar:
Add [harper-ls](https://github.com/automattic/harper) to the `vim.lsp` module.

- Style and consistency
- Package(s) built:
  - [x] `.#maximal`
- Tested on platform(s)
  - [x] `x86_64-linux`
  - [ ] `aarch64-linux`
  - [ ] `x86_64-darwin`
  - [ ] `aarch64-darwin`

---

### DIFF
--- a/modules/plugins/lsp/harper-ls/harper-ls.nix
+++ b/modules/plugins/lsp/harper-ls/harper-ls.nix
@@ -1,0 +1,35 @@
+{lib, ...}: let
+  inherit (lib.options) mkOption mkEnableOption;
+  inherit (lib.types) anything attrsOf;
+in {
+  options.vim.lsp.harper-ls = {
+    enable = mkEnableOption "Harper grammar checking LSP";
+    settings = mkOption {
+      type = attrsOf anything;
+      default = {};
+      example = {
+        userDictPath = "";
+        workspaceDictPath = "";
+        fileDictPath = "";
+        linters = {
+          BoringWords = true;
+          PossessiveNoun = true;
+          SentenceCapitalization = false;
+          SpellCheck = false;
+        };
+        codeActions = {
+          ForceStable = false;
+        };
+        markdown = {
+          IgnoreLinkTitle = false;
+        };
+        diagnosticSeverity = "hint";
+        isolateEnglish = false;
+        dialect = "American";
+        maxFileLength = 120000;
+        ignoredLintsPath = {};
+      };
+      description = "Settings to pass to harper-ls";
+    };
+  };
+}
--- a/configuration.nix
+++ b/configuration.nix
@@ -31,6 +31,7 @@ isMaximal: {
       lspSignature.enable = !isMaximal; # conflicts with blink in maximal
       otter-nvim.enable = isMaximal;
       nvim-docs-view.enable = isMaximal;
+      harper-ls.enable = isMaximal;
     };
 
     debugger = {
--- a/docs/release-notes/rl-0.8.md
+++ b/docs/release-notes/rl-0.8.md
@@ -524,3 +524,9 @@
 [diced](https://github.com/diced):
 
 - Fixed `typescript` treesitter grammar not being included by default.
+
+[gmvar](https://github.com/gmvar):
+
+[harper-ls]: https://github.com/Automattic/harper
+
+- Add [harper-ls] to the `vim.lsp` module.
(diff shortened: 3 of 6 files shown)


modules/plugins/lsp/harper-ls/harper-ls.nix
```diff
@@ -0,0 +1,35 @@
+{lib, ...}: let
+  inherit (lib.options) mkOption mkEnableOption;
+  inherit (lib.types) anything attrsOf nullOr;
+in {
+  options.vim.lsp.harper-ls = {
+    enable = mkEnableOption "Harper grammer checking LSP";
+    settings = mkOption {
+      type = nullOr (attrsOf anything);
+      default = null;
```
horriblename: is null necessary? passing empty table `{}` to `vim.lsp.servers.harper-ls.settings` should be fine
```suggestion
      type = attrsOf anything;
      default = {};
```

horriblename (changes requested):

modules/plugins/lsp/harper-ls/harper-ls.nix
gmvar: Done.

modules/plugins/lsp/harper-ls/harper-ls.nix
```diff
@@ -0,0 +1,35 @@
...
+        },
+        markdown = {
+          IgnoreLinkTitle = false
+        },
+        diagnosticSeverity = "hint",
+        isolateEnglish = false,
+        dialect = "American",
+        maxFileLength = 120000,
+        ignoredLintsPath = {}
+      '';
```
horriblename: @NotAShelf @Soliprem should this section be removed in favor of Configuring `vim.lsp.servers.harper-ls.settings` directly? we removed the settings aliases from language modules since we couldn't make it work with multiple servers but maybe this one is fine?

NotAShelf: We should be consistent with our language module options so that their behaviour is intuitive. If a language module provides, say, settings while others don't then we risk confusing the user.

gmvar: ### `harper-ls` is an LSP server, not a language or nvim plugin

Should that rule still follow even though this isn't a language module but a config option for a specific LSP? 

I have only briefly taken a look at `mkPluginSetupOption`, but I'm not sure if that is the correct path to take. The issue is that any config for `harper-ls` needs to generate the config to match this structure.
```lua
vim.lsp.config["harper-ls"] = {
  ["settings"] = {
    ["harper-ls"] = {
      ["foo"] = "bar" -- what the user defines in vim.lsp.servers.harper-ls.settings
    }
  }
}
```

and not:
```lua
require("harper-ls").setup({["foo"] = "bar"})
```

NotAShelf: harper-ls is in a weird case, and while it is, in fact, not a language module only language modules have relied on LSP functionality in nvf to this date. Thus harper especially annoying to categorize because:

1. It does rely on an internal API (vim.lsp) so we'd be wrapping Neovim's own API
2. It is not a set-and-forget plugin and therefore sticking it into something like vim.utility would be ugly as only plugins go under those categories.

That said, I think we can make a special case for LSPs of such nature (I believe there other, similar LSPs for e.g. AI functionality) taking `settings` so there would be some resemblance of future-proofing when you think about it. I also think a small refactor is applicable to how we vendor LSP and formatter configurations.

*P.S.: I hope you are not implying a lack in my eyesight (lest intelligence) with the bold text. I'm sleepy, not blind. Would be rude of you to do so.*

modules/plugins/lsp/harper-ls/harper-ls.nix
```diff
@@ -0,0 +1,35 @@
...
```
NotAShelf: The type is `attrsOf anything` but the example uses a (multiline) string with Lua code. Should be the Nix attribute set that will be converted, not the conversion results.

gmvar: Thanks for catching that, fixed.

modules/plugins/lsp/harper-ls/harper-ls.nix
gmvar: I apologize if my formatting came across as rude; that was not my intention at all.

PR #1178, opened 2025-10-10
lib: rewrite toLuaObject, deprecate everything else

Gerg-L:
functionally a no-op


horriblename:
did you replace all instances of expToLua with tLuaObject? iirc there are a lot more of those

Gerg-L:
That should be actually good now, and ci shouldn't error

PR #1185, opened 2025-10-15
languages/rust: re-apply crates.nvim changes

NotAShelf:
This got reverted down the line, which makes me a little concerned about what else could've gotten reverted during my failed rebase.

Adds back changes that somehow got reverted, by backtracking my previous PR.


horriblename (approved):

horriblename:
seems like this isn't needed any more

PR #1190, opened 2025-10-17
lib: don't duplicate nixpkgs' lib

alfarelcynthesis:
The current implementation of nvf's custom lib uses nixpkgs' `lib.extend` function.

This is [intended as an escape hatch for buggy behaviour in nixpkgs](https://github.com/hsjobeki/nixpkgs/blob/484a0477d83c9369af0213b83d08a52e4a12f327/lib/default.nix#L32) and adds unnecessary eval time.
As well, when I was initially looking at nvf's codebase, the use of the `lib.nvim` namespace obfuscated where functions were actually coming from in nvf's code, but this might be a nonissue depending on personal opinion.

This PR converts the codebase to a separate extensible fixed-point, implemented similarly to nixpkgs' lib, without being a direct extension of it.

This is fairly simple and imo actually reduces complexity with not having to keep around the extension function.
I think it's fair to call this a reduction in technical debt.

The current implementation is incomplete, since I want to know whether this might be accepted and, if so, whether to take the approach:

1. With a minimal diff that keeps the reduced eval time and prevents any the issues nixpkgs warns against, but doesn't improve code clarity (by separating the two libs explicitly).
   This means `_module.args.lib = pkgs.lib // {nvim = self.lib;}`, probably.
2. Which does all, with treewide changes from `lib.nvim` -> `self.lib`.

## Sanity Checking

  - This shouldn't change anything for users, since the version exposed by the flake already stripped nixpkgs out.
- [ ] I have tested, and self-reviewed my code
  - Reviewed a previous version which worked without issues, want advice before proceeding here and doing more thorough testing.
- Style and consistency
  - [x] My changes are consistent with the rest of the codebase (in terms of style)
    - No need for this to be breaking.
- Package(s) built:
  - [ ] `.#maximal`
- Tested on platform(s)
  - [ ] `x86_64-linux`
  - [ ] `aarch64-linux`
  - [ ] `x86_64-darwin`
  - [ ] `aarch64-darwin`

---

### DIFF
--- a/modules/default.nix
+++ b/modules/default.nix
@@ -1,7 +1,7 @@
 {
   self,
   inputs,
-  lib,
+  ...
 }: {
   pkgs,
   extraSpecialArgs ? {},
@@ -10,17 +10,20 @@
   extraModules ? [],
   configuration ? {},
 }: let
+  inherit (pkgs) lib;
+  inherit (lib.modules) evalModules;
   inherit (lib.strings) toString;
-  inherit (lib.lists) concatLists;
+  inherit (lib.trivial) warn;
+  inherit (lib.lists) concatLists optional optionals;
 
-  # import modules.nix with `check`, `pkgs` and `lib` as arguments
+  # import modules.nix with `check` and `pkgs` as arguments
   # check can be disabled while calling this file is called
   # to avoid checking in all modules
-  nvimModules = import ./modules.nix {inherit pkgs lib;};
+  nvimModules = import ./modules.nix {inherit pkgs;};
 
   # evaluate the extended library with the modules
   # optionally with any additional modules passed by the user
-  module = lib.evalModules {
+  module = evalModules {
     specialArgs =
       extraSpecialArgs
       // {
@@ -30,12 +33,12 @@
     modules = concatLists [
       nvimModules
       modules
-      (lib.optional (configuration != {}) (lib.warn ''
+      (optional (configuration != {}) (warn ''
           nvf: passing 'configuration' to lib.neovimConfiguration is deprecated.
         ''
         configuration))
 
-      (lib.optionals (extraModules != []) (lib.warn ''
+      (optionals (extraModules != []) (warn ''
           nvf: passing 'extraModules' to lib.neovimConfiguration is deprecated, use 'modules' instead.
         ''
         extraModules))


alfarelcynthesis:
(@NotAShelf since I want opinions before marking this ready for (final) review)

NotAShelf:
To be honest I'm not a huge fan of separating lib and nvf-lib, but I can see why one would want to. I'd like to defer to @horriblename as I'm currently too occupied irl to deal with large-scale refactors myself.

horriblename:
I am completely out of my depth with this, the link gives some very scary warnings but gives no explanation on why, so I'm not really sure what to do with that info

If you can help me understand any of this it would be much appreciated :)

alfarelcynthesis:
The currently-used `nixpkgs.lib.extend` function computes `lib` with the provided overlay function (in `stdlib-extended.nix`), in the same way a `nixpkgs` overlay works.

This PR converts the custom library to be created in the same way as `lib` itself [is created](https://github.com/NixOS/nixpkgs/blob/0c3a6301677a05013ec1560a7015e69e3e3c3f5b/lib/default.nix#L40), as a fixed point containing only the local library functions.

___

The warnings they mention probably aren't really relevant, except for if someone using `nvf` makes their own extended `lib` and tries to provide it using `extraSpecialArgs` to the (standalone) module.
I'm honestly not sure if that would produce a reasonable eval error or if it would yell about missing `lib.nvim` functions (if the user `lib` overrides the `nvf`-provided `specialArg`).
Essentially, if `lib.extend` (and probably `specialArgs`/`extraSpecialArgs`) are used in multiple places it can cause issues, but I'm not sure how important that actually is.

I assume they also warn against modifying the value `lib` since people expect the variable to behave in a certain way.
The current design doesn't directly punish that assumption (since it only adds things), so it isn't necessarily bad.

- In that vein, in my opinion it would be easier to reason about the project (especially as an occasional contributor) if `lib` was same thing as it is everywhere else and custom functionality come from `self` or another argument representing this project.
- I also think that sticking to using `lib` as the `nixpkgs` maintainers expect is just a generally good idea, which is why I linked the warning.
  Sorry if it was confusing, I should have been clearer about that.

Ease of understanding is (mostly) a question of maintainer preference though, so if y'all don't actually see benefits it probably isn't worth doing.

___

The eval benefits of separating the two library evaluations are untested but probably very slightly positive.
That isn't enough to make this worth it on its own, though.

___

Let me know if I talked about the wrong things or if you want changes/to close this :)

(Sorry it took me a couple weeks to get back to this, I was busy and figured this wasn't particularly urgent)

alfarelcynthesis:
This probably isn't worth reviewing at this point, but thank you for at least being open to my slightly deranged suggestion.

This was originally motivated by figuring out a system for configurable keymap option defaults, which this made a bit nicer in my head.
I'll hopefully get around to implementing a small-scope and reasonable-to-review PR for the keymap option stuff soon enough, but separately.

NotAShelf:
Thank you for your time and concern ❤️

PR #1249, opened 2025-11-27
feat(autocompletion): allow blink-cmp package override

blackxored:
blink-cmp is broken because of some rust flags that need to be passed in (See https://github.com/NixOS/nixpkgs/pull/464350). Had to fork so it would be nice if you could override the package as in this PR.

### DIFF
--- a/modules/plugins/completion/blink-cmp/blink-cmp.nix
+++ b/modules/plugins/completion/blink-cmp/blink-cmp.nix
@@ -1,5 +1,5 @@
-{lib, ...}: let
-  inherit (lib.options) mkEnableOption mkOption literalMD;
+{lib, pkgs, ...}: let
+  inherit (lib.options) mkEnableOption mkOption mkPackageOption literalMD;
   inherit (lib.types) bool listOf str either attrsOf submodule enum anything int nullOr;
   inherit (lib.nvim.types) mkPluginSetupOption luaInline pluginType;
   inherit (lib.nvim.binds) mkMappingOption;
@@ -29,6 +29,7 @@
 in {
   options.vim.autocomplete.blink-cmp = {
     enable = mkEnableOption "blink.cmp";
+    package = mkPackageOption pkgs ["vimPlugins" "blink-cmp"] { nullable = true; };
     setupOpts = mkPluginSetupOption "blink.cmp" {
       sources = {
         default = mkOption {
--- a/modules/plugins/completion/blink-cmp/config.nix
+++ b/modules/plugins/completion/blink-cmp/config.nix
@@ -17,7 +17,9 @@
   inherit (cfg) mappings;
 
   getPluginName = plugin:
-    if typeOf plugin == "string"
+    if cfg.package != null
+      then cfg.package.pname
+    else if typeOf plugin == "string"
     then plugin
     else if (plugin ? pname && (tryEval plugin.pname).success)
     then plugin.pname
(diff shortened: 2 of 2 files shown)


NotAShelf:
Okay, I'm ammendable to such a change but there are two crucial points to consider.

1. We try to avoid relying on Nixpkgs for plugins. They are maintained haphazardly at best, and I do not wish to deal with upstream. Currently blink is managed by flake packages locally, and finding a way to override that one would be better.

2. If we are adding one package option, then we need to support the package option for other plugin modules as well. This has been something I've been considering, but since most plugins are just fetch + unpack + install, overriding the pin (as documented) is preferable.

An option that addresses both of those points might be to allow overriding built packages instead of just npins sources. This way you could change what the package corresponds to and thus use any packaged plugin.

blackxored:
Good points all-round, I just need a way to override if there's such a broken state I cannot rebuild until this project gets updated (the missing Rust flags being one example), and also blink-cmp has been on 1.6.0 for a while so it was Nix that broke it.

alfarelcynthesis:
`pluginOverrides` works for some things, idk if it would for blink.

alapshin:
I was able to override blink-cmp package without using package from nixpkgs using approach like this

```
{
  lib,
  pkgs,
  inputs,
  ...
}:
let
  inherit (lib.generators) mkLuaInline;
  system = pkgs.stdenv.hostPlatform.system;
  packages = inputs.nvf.packages."${system}";
in
{
  home.sessionVariables.EDITOR = "nvim";
  programs.nvf = {
    enable = true;
    settings.vim = {
      viAlias = true;
      vimAlias = true;
      enableLuaLoader = true;

      pluginOverrides = {
        blink-cmp = packages.blink-cmp.overrideAttrs (oldAttrs: {
          env = oldAttrs.env // {
            RUSTFLAGS = lib.optionalString pkgs.stdenv.hostPlatform.isDarwin "-C link-arg=-undefined -C link-arg=dynamic_lookup";
          };
        });
      };
    };
  };
}
```

One have to pass nvf input and copy-paste RUSTFLAGS from nixpkgs PR. Guess a proper way to fix it would be update package definition for blink-cmp https://github.com/NotAShelf/nvf/blob/v0.8/flake/pkgs/by-name/blink-cmp/package.nix and add missing RUSTFLAGS.

blackxored:
> I was able to override blink-cmp package without using package from nixpkgs using approach like this
> 
> ```
> {
>   lib,
>   pkgs,
>   inputs,
>   ...
> }:
> let
>   inherit (lib.generators) mkLuaInline;
>   system = pkgs.stdenv.hostPlatform.system;
>   packages = inputs.nvf.packages."${system}";
> in
> {
>   home.sessionVariables.EDITOR = "nvim";
>   programs.nvf = {
>     enable = true;
>     settings.vim = {
>       viAlias = true;
>       vimAlias = true;
>       enableLuaLoader = true;
> 
>       pluginOverrides = {
>         blink-cmp = packages.blink-cmp.overrideAttrs (oldAttrs: {
>           env = oldAttrs.env // {
>             RUSTFLAGS = lib.optionalString pkgs.stdenv.hostPlatform.isDarwin "-C link-arg=-undefined -C link-arg=dynamic_lookup";
>           };
>         });
>       };
>     };
>   };
> }
> ```
> 
> One have to pass nvf input and copy-paste RUSTFLAGS from nixpkgs PR. Guess a proper way to fix it would be update package definition for blink-cmp https://github.com/NotAShelf/nvf/blob/v0.8/flake/pkgs/by-name/blink-cmp/package.nix and add missing RUSTFLAGS.

I'll try that in a bit, thanks! Aren't you resetting RUSTFLAGS on non-darwin though? Should probably move up the chain with `lib.optionalAttrs`.

NotAShelf:
The breakage should've been fixed on latest v0.8 commit.

Issue #1195, opened 2025-10-19
feat(0.8): Move language-specific plugins into language extensions


sjcobb2022:
### I have verified that this feature request has not been made before

- [x] I have checked the [issues tab](https://github.com/notashelf/nvf/issues?q=is%3Aissue), and did not find a similar feature request. I understand that my issue will be closed if it is a duplicate.

### Feature Type

Other

### Feature description

Proposal
- Default to configuring language servers with vim.lsp.config.
- Provide an opt-in `vim.languages.<lang>.extensions` to let language-specific plugins (e.g. rustaceanvim, haskell-tools.nvim) take over LSP management by disabling the built-in enable for that server. Similar to how we have done it with `crates.nvim`

Example
```nix
{
  vim.languages.haskell.enable = true;
  vim.languages.haskell.extensions.haskell-tools-nvim.enable = true;

  vim.languages.rust.enable = true;
  vim.languages.rust.extensions = {
    rustaceanvim.enable = true;
    crates-nvim.enable = true;
  };
}
```

Notes
- When an extension is enabled, ensure the corresponding `vim.lsp.enable` is disabled to avoid double-registration. (breaks things)

Compatibility references
- https://github.com/mrcjkb/haskell-tools.nvim/commit/53d9888aebe27cd50c0404d88a7aa51d7341784a
- https://github.com/mrcjkb/rustaceanvim/pull/660

NotAShelf:
I think this could be a valuable addition. The requirement of plugins to set up basic language modules has been nagging me for a bit, and moving to base `vim.lsp` and enabling rustaceanvim & friends (both of which are *very* invasive imo) could be the way to go. Though I'd like to discuss this a bit further.

Namely, I want to debate what we'd lose by making those disabled by default (since extensions *are* disabled by default) and the discoverability of those options. It is often times the case where users don't notice an option, and remain blisfully unaware of nvf's capabilities.

horriblename:
rustaceanvim and haskell-tools.nvim are fairly advanced tools, and offers much more that LSP/DAP. haskell-tools explicitly states that new-comers should use lspconfig for hls config, and only advanced users need haskell-tools. Anyone looking for advanced features (hopefully) knows the plugin's existence and will check for themselves. Everyone else using LSP only isn't even going to realize haskell-tools absence.

I'm not sure if this applies for all language plugins though

sjcobb2022:
[...]

PR #1271, opened 2025-12-10
languages/csharp: add razor support

### LINKED ISSUE #1268: roslyn-ls and rz-ls for csharp
### I have verified that this feature request has not been made before

- [x] I have checked the [issues tab](https://github.com/notashelf/nvf/issues?q=is%3Aissue), and did not find a similar feature request. I understand that my issue will be closed if it is a duplicate.

### Feature Type

Other

### Feature description

It would be great to have roslyn-ls and rzls support for C# (razor development) as an alternative to csharp-ls.

### Alternatives

_No response_

### Additional Context

_No response_

CaueAnjos:
It adds the **roslyn-ls** configuration and the **roslyn-nvim** plugin. Razor support works _partially_. The roslyn-nvim team announced that they now support Razor, but I think the version of roslyn-ls available in nixpkgs isn’t up to date. Because of that, Razor files may still produce errors. Aside from this, everything works fine.

This PR should fix #1268

## Sanity Checking

- Style and consistency
  - [x] _(For breaking changes)_ I have included a migration guide
- Package(s) built:
  - [x] `.#maximal`
- Tested on platform(s)
  - [x] `x86_64-linux`
  - [ ] `aarch64-linux`
  - [ ] `x86_64-darwin`
  - [ ] `aarch64-darwin`

---

### DIFF
--- a/modules/plugins/languages/csharp.nix
+++ b/modules/plugins/languages/csharp.nix
@@ -24,10 +24,8 @@
   in
     optionalString (key != null) "vim.keymap.set('n', '${key}', ${action}, {buffer=bufnr, noremap=true, silent=true, desc='${desc}'})";
 
-  # Omnisharp doesn't have colors in popup docs for some reason, and I've also
-  # seen mentions of it being way slower, so until someone finds missing
-  # functionality, this will be the default.
-  defaultServers = ["csharp_ls"];
+  # NOTE: roslyn is the most feature-rich option, and its Razor integration is better than csharp-ls (which only supports .cshtml).
+  defaultServers = ["roslyn"];
   servers = {
     omnisharp = {
       cmd = mkLuaInline ''
@@ -118,8 +116,8 @@
     };
 
     csharp_ls = {
-      cmd = [(lib.getExe pkgs.csharp-ls)];
-      filetypes = ["cs"];
+      cmd = [(lib.getExe pkgs.csharp-ls) "--features" "razor-support"];
+      filetypes = ["cs" "razor"];
       root_dir = mkLuaInline ''
         function(bufnr, on_dir)
           local function find_root_pattern(fname, lua_pattern)
@@ -137,17 +135,28 @@
       };
     };
 
-    roslyn_ls = {
+    roslyn = let
+      pkg = pkgs.vscode-extensions.ms-dotnettools.csharp;
+      pluginRoot = "${pkg}/share/vscode/extensions/ms-dotnettools.csharp";
+      exe = "${pluginRoot}/.roslyn/Microsoft.CodeAnalysis.LanguageServer";
+      razorSourceGenerator = "${pluginRoot}/.razorExtension/Microsoft.CodeAnalysis.LanguageServer";
+      razorDesignTimePath = "${pluginRoot}/.razorExtension/Targets/Microsoft.NET.Sdk.Razor.DesignTime.targets";
+      razorExtension = "${pluginRoot}/.razorExtension/Microsoft.VisualStudioCode.RazorExtension.dll";
+    in {
       cmd = mkLuaInline ''
         {
-          ${toLuaObject (getExe pkgs.roslyn-ls)},
-          '--logLevel=Warning',
-          '--extensionLogDirectory=' .. vim.fs.dirname(vim.lsp.get_log_path()),
-          '--stdio',
+          "dotnet",
+          "${exe}.dll",
+          "--stdio",
+          "--logLevel=Information",
+          "--extensionLogDirectory=" .. vim.fs.dirname(vim.lsp.get_log_path()),
+          "--razorSourceGenerator=${razorSourceGenerator}",
+          "--razorDesignTimePath=${razorDesignTimePath}",
+          "--extension=${razorExtension}",
         }
       '';
 
-      filetypes = ["cs"];
+      filetypes = ["cs" "razor"];
       root_dir = mkLuaInline ''
         function(bufnr, on_dir)
           local function find_root_pattern(fname, lua_pattern)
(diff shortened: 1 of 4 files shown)


NotAShelf:
Could you target the v0.8 branch? The LSP experience was mostly overhauled and improved, so this PR will likely benefit from the new API.

CaueAnjos:
I've noticed a lot has changed. I’m working on it! 😁

CaueAnjos:
I am having some trouble figuring out where to add the roslyn-ls path to `vim.env.PATH`. roslyn-nvim needs this so it can find roslyn-ls. Where should I set it?

Soliprem:
we merged 0.8 into main! So you can target main directly, now (sorry for the double switch)

CaueAnjos:
I figured it out! It wasn't a roslyn-ls version problem. Razor wasn't working correctly because the .razorExtension wasn't being mapped. Now everything works perfectly.

poseidon-rises:
I actually added roslyn-ls in v8.0 already, you should be able to use it now that's pulled into main.

CaueAnjos:
> I actually added roslyn-ls in v8.0 already, you should be able to use it now that's pulled into main.

I noticed. It’s a pretty good implementation. I’ve just added Razor support. Your implementation doesn’t have that, and I need it for Blazor projects.

modules/plugins/languages/csharp.nix
```diff
@@ -195,11 +151,81 @@ in {
...
     (mkIf cfg.treesitter.enable {
       vim.treesitter.enable = true;
-      vim.treesitter.grammars = [cfg.treesitter.package];
+      vim.treesitter.grammars = [cfg.treesitter.csPackage cfg.treesitter.razorPackage];
+    })
+
+    (mkIf (cfg.lsp.enable && lib.elem "roslyn" cfg.lsp.servers) {
+      vim.luaConfigRC.roslyn-path = ''
+        -- NOTE: this is required by roslyn-nvim to find roslyn
+        vim.env.PATH = vim.env.PATH .. ":${pkgs.roslyn-ls}/bin"
```
horriblename: 1. add the package to `vim.extraPackages` instead
~~2. the package needs to be overridable somehow, preferably with `vim.lsp.servers.roslyn.cmd`, or if it's really not doable, add an option  `vim.languages.csharp.lsp.package` (this deviates from our usual pattern so an appropriate warning in case `cmd` is used would be appreciated)~~ doesn't really apply if we wait for the nixpkgs PR

modules/plugins/languages/csharp.nix
```diff
@@ -195,11 +151,81 @@ in {
...
+            local mason = expanded_mason == "$MASON"
+                and vim.fs.joinpath(vim.fn.stdpath("data"), "mason")
+                or expanded_mason
+            local mason_packages = vim.fs.joinpath(mason, "packages")
+
+            -- Define target paths
+            local targets = {
+                vim.fs.joinpath(mason_packages, "roslyn", "libexec", ".razorExtension"),
+                vim.fs.joinpath(mason_packages, "roslyn-unstable", "libexec", ".razorExtension"),
+            }
```
horriblename: use `vim.fn.stdpath("data") .. "/mason"` instead of hard-coding `~/.local/share`

I don't really like this approach overall. I would rather wait until `https://github.com/NixOS/nixpkgs/issues/472357` is merged. Will probably need to patch out the warning in roslyn-nvim unless they start supporting it themselves though

modules/plugins/languages/csharp.nix
```diff
@@ -120,65 +106,35 @@
...
+      filetypes = ["cs" "razor" "cshtml"];
+      root_marks = [".sln" ".csproj"];
     };
   };
 
   extraServerPlugins = {
     omnisharp = ["omnisharp-extended-lsp-nvim"];
     csharp_ls = ["csharpls-extended-lsp-nvim"];
-    roslyn_ls = [];
+    roslyn = ["roslyn-nvim"];
```
horriblename, line 179: this is technically a breaking change, a warning about removal of "roslyn_ls" in `lsp.servers` would be nice

horriblename (commented):
Thanks, but overall, I think we should wait for the nixpkgs PR to land first before adding razor support (see my other comments)

CaueAnjos:
> I don't really like this approach overall. I would rather wait until `https://github.com/NixOS/nixpkgs/issues/472357` is merged. Will probably need to patch out the warning in roslyn-nvim unless they start supporting it themselves though

I’ve rethought the symlink approach, and I now believe this solution is more reproducible and, overall, more stable.
The code is also prepared for the future .razorExtension bundled with roslyn-ls.

horriblename:
is this ready for review now?

modules/plugins/languages/csharp.nix
```diff
@@ -120,65 +106,65 @@
...
-          local function find_root_pattern(fname, lua_pattern)
-            return vim.fs.root(0, function(name, path)
-              return name:match(lua_pattern)
-            end)
-          end
-
-          local fname = vim.api.nvim_buf_get_name(bufnr)
-          on_dir(find_root_pattern(fname, "%.sln$") or find_root_pattern(fname, "%.csproj$"))
-        end
-      '';
```
horriblename, line 134: the original root_dir function and the new root_marks are not equivalent, the root_marks do not match `ProjectName.csproj` and `ProjectName.sln`, don't change these

modules/plugins/languages/csharp.nix
```diff
@@ -188,18 +174,43 @@ in {
...
   config = mkIf cfg.enable (mkMerge [
     (mkIf cfg.treesitter.enable {
       vim.treesitter.enable = true;
-      vim.treesitter.grammars = [cfg.treesitter.package];
+      vim.treesitter.grammars = [cfg.treesitter.csPackage cfg.treesitter.razorPackage];
     })
 
+    (mkIf (cfg.lsp.enable && lib.elem "roslyn" cfg.lsp.servers)
+      {
+        vim.extraPackages = [pkgs.roslyn-ls pkgs.vscode-langservers-extracted];
```
horriblename: Don't add the LSP package to PATH, set the cmd for roslyn directly with `vim.lsp.servers.*.cmd`

modules/plugins/languages/csharp.nix
```diff
@@ -188,18 +174,43 @@ in {
...
+
+        # NOTE: roslyn-nvim requires razorExtension dll's
+        # See: https://github.com/seblyng/roslyn.nvim/blob/main/lua/roslyn/health.lua
+        vim.luaConfigRC.razorExtension = ''
+          vim.env.MASON = "${razorExtensions.${cfg.lsp.razorExtension}}"
```
horriblename: I don't think abusing $MASON like this is a good idea. What if some other plugin sees $MASON and starts doing mason things it's not supposed to? What if another LSP needs this hack as well?

Adapting from <https://github.com/NixOS/nixpkgs/issues/472357#issuecomment-3752734454>, something like this should work (untested):

```nix
let
  pkg = pkgs.vscode-extensions.ms-dotnettools.csharp;
  pluginRoot =  "${pkg}/share/vscode/extensions/ms-dotnettools.csharp";
  exe ="${pluginRoot}/.roslyn/Microsoft.CodeAnalysis.LanguageServer":
  razorSourceGenerator = "${pluginRoot}/.razorExtension/Microsoft.CodeAnalysis.LanguageServer":
  razorDesignTimePath = "${pluginRoot}/.razorExtension/Targets/Microsoft.NET.Sdk.Razor.DesignTime.targets";
  razorExtension = "${pluginRoot}/.razorExtension/Microsoft.VisualStudioCode.RazorExtension.dll";

in {
  cmd = mkLuaInline ''
    {
      "${exe}",
      "--stdio",
      "--logLevel=Information",
      "--extensionLogDirectory=" .. vim.fs.dirname(vim.lsp.get_log_path()),
      "--razorSourceGenerator=${razorSourceGenerator}",
      "--razorDesignTimePath=${razorDesignTimePath}",
      "--extension=${razorExtension}",
    }
  '';
}

```

set vim.lsp.servers.roslyn.cmd to the above

CaueAnjos: This solution works perfectly. The only required adjustment is to prepend `dotnet` before `"${exe}"`. Without this, `roslyn-ls` incorrectly reports that the .NET SDK is missing, even though it is installed.

There is one additional observation: I tested Razor support in `csharp-ls`, and while it does function, the implementation is quite minimal. It only recognizes `.cshtml` files and offers limited functionality beyond basic support. Nevertheless, I decided to include it.

modules/plugins/languages/csharp.nix
CaueAnjos: Now I warn it on `lsp.servers`. Just not sure if I did it the right way.

horriblename: sorry for the delay,

following some discussion in #1195, we've decided to avoid depending on plugins for basic LSP support, could you please:
1. add the `roslyn` server as a separate entry, and keeping the old `roslyn-ls`
2. set the default back to roslyn-ls
3. add a dedicated `roslyn-nvim` option as `vim.languages.csharp.extensions.roslyn-nvim`, in the same style as [crates-nvim](https://github.com/NotAShelf/nvf/blob/6681e33727409d4ccfa687de981b594110a735d6/modules/plugins/languages/rust.nix#L106)

the deprecation msg is then not needed anymore

CaueAnjos:
I couldn’t update the original branch because GitHub Actions deleted it, so I created a new pull request (#1498) with the updated changes.

Sorry for the inconvenience.

PR #1294, opened 2025-12-16
utility/ccc: add setupOpts option

jfeo:
Added `ccc.nvim` options `vim.utility.ccc.inputs` and `vim.utility.ccc.outputs` to make input color systems and output color formats configurable.

Notes:
1. The change should be backwards compatible since the previous hard-coded values are now the option defaults.
2. The option descriptions are more or less copied from `ccc.nvim`, with minor modifications and typo fixes.
3. I am adding release notes to `rl-0.8.md`, is that correct? I suppose it adds to the current release
4. I added a colon to the previous release note entry, is that alright?
5. I did not use the `toLuaObject` function as it would quote the list values. Not sure if there is a better way?

## Sanity Checking

- Style and consistency
- Package(s) built:
  - [x] `.#maximal`
- Tested on platform(s)
  - [x] `x86_64-linux`
  - [ ] `aarch64-linux`
  - [ ] `x86_64-darwin`
  - [ ] `aarch64-darwin`

---

### DIFF
--- a/modules/plugins/utility/ccc/ccc.nix
+++ b/modules/plugins/utility/ccc/ccc.nix
@@ -1,12 +1,6 @@
 {lib, ...}: let
   inherit (lib.options) mkOption mkEnableOption;
-  inherit
-    (lib.types)
-    anything
-    attrsOf
-    listOf
-    enum
-    ;
+  inherit (lib.types) anything attrsOf listOf enum;
   inherit (lib.nvim.binds) mkMappingOption;
   inherit (lib.nvim.types) mkPluginSetupOption luaInline;
   inherit (lib.generators) mkLuaInline;


Soliprem:
the changes look fine, however we're trying to move as many plugins as possible to the freeform `toLuaOption`. I'll @NotAShelf since this is more of a "general direction" decision.

Would you feel comfortable with giving the refactor a shot?

jfeo:
> the changes look fine, however we're trying to move as many plugins as possible to the freeform `toLuaOption`. I'll @NotAShelf since this is more of a "general direction" decision.
> 
> Would you feel comfortable with giving the refactor a shot?

I am up for that -- might as well do it properly. But as I am a newbie both with nix and nvf this simple change results in a few questions :) Hope that's ok!

After reading the docs and code a bit more thoroughly I found `mkLuaInline` and `mkPluginSetupOption`. I could make a change to something like the following, I think:

```nix
// ...utility/ccc/ccc.nix
options.vim.utility.ccc = {
  // ...
  setupOpts = mkPluginSetupOption "ccc.nvim" {
    inputs = mkOption { /* ... */ };
    outputs = mkOption { /* ... */ };
  };
  // ...
};

// ...utility/ccc/config.nix
vim.pluginRC.ccc = entryAnywhere ''
  local ccc = require("ccc")
  ccc.setup(${toLuaObject (mapCccSetupOptions cfg.setupOpts)})
''
```
where I think something like `mapCccSetupOptions` would be needed to map options to lua literals, e.g. output `css_rgb` to `mkLuaInline "ccc.output.css_hsl"`

However then I would have to decide if the existing `options.vim.utility.ccc.mappings` should be migrated. It seems that currently they are not used to actually write any mappings at all, which is a separate issue.

jfeo:
@Soliprem 

I updated the PR with a refactoring using `mkPluginSetupOption`, with suboptions for each of the options that existed in the original hard-coded configuration.

I think this is the most sensible way to achieve a default configuration equal to the original, since it does not seem like `mkPluginSetupOption` can accept a default configuration.

I think I should rename the PR as well, if you decide to accept there changes :)

modules/plugins/utility/ccc/ccc.nix
```diff
@@ -1,10 +1,154 @@
...
+      highlighter = mkOption {
+        type = luaInline;
+        default = mkLuaInline ''
+          {
+            auto_enable = true,
+            max_byte = 2 * 1024 * 1024, -- 2mb
+            lsp = true,
+            filetypes = colorPickerFts,
+          }
+        '';
```
NotAShelf: As far as I'm aware `higlighter` is a Lua table, which the setupOpts API suppots in conversion. Any reason this oculdn't be an attribute set?

`attrsOf anyting` strikes me as the correct type here; any Lua values (such as `colorPickerFts`) can be deliberately inlined with `mkLuaInline instead of the whole option.

modules/plugins/utility/ccc/ccc.nix
```diff
@@ -1,10 +1,154 @@
...
+        default = mkLuaInline ''
+          {
+            ccc.picker.hex,
+            ccc.picker.css_rgb,
+            ccc.picker.css_hsl,
+            ccc.picker.ansi_escape {
+              meaning1 = "bright", -- whether the 1 means bright or yellow
+            },
+          }
+        '';
```
NotAShelf: This can probably be converted from a Nix list, but I'm a little unhappy about how it would look. We can leave this one as is.

modules/plugins/utility/ccc/ccc.nix
```diff
@@ -1,10 +1,154 @@
...
+        description = ''
+          List of formats that can be detected by |:CccPick| to be activated.
```
NotAShelf:
```suggestion
          List of formats that can be detected by {command}`:CccPick` to be activated.
```

NDG doesn't have a way of processing Neovim's help text yet, so it's nicer to parse this as a command role. In the future I might add a postprocessor that links the appropriate Neovim documentation from `|:cmd|`.

modules/plugins/utility/ccc/ccc.nix
```diff
@@ -1,10 +1,154 @@
...
+          UI is opened. "show" and "hide" mean as they are. "auto" makes the
+          slider appear only when the alpha value can be picked up.
+        '';
+      };
+
+      recognize = mkOption {
+        type = luaInline;
+        default = mkLuaInline ''
+          { output = true }
+        '';
```
NotAShelf: This could be either a list (so it converts to a table) or `luaInline`. I find `luaInline` to be generally confusing, so it *might* be nicer to go ahead and support lists *anyway* but it's not a blocker.

modules/plugins/utility/ccc/config.nix
```diff
@@ -5,48 +5,23 @@
...
+  inherit (lib.nvim.lua) toLuaObject;
+  inherit (lib.generators) mkLuaInline;
   cfg = config.vim.utility.ccc;
+  mkLuaIdentifier = prefix: identifier: mkLuaInline "${prefix}${identifier}";
+  mapSetupOptions = setupOpts:
+    setupOpts
+    // {
+      inputs = map (mkLuaIdentifier "ccc.input.") setupOpts.inputs;
+      outputs = map (mkLuaIdentifier "ccc.output.") setupOpts.outputs;
+    };
```
NotAShelf: This is pretty confusing for the user, and it's hidden behaviour as far as they're concerned. Could we not check the setupOpts table instead for "broken" values that *need* to be caught?

We could also skip this altogether and expect thte user to provide the prefix themselves, which is how the plugin is meant to be used without the Nix abstraction layer.

jfeo: I can see that it is desireable to avoid specific behaviour like this for various plugins :)

I removed this, so that no hidden mapping occurs, by changing the type of the option to `listOf luaInline`. The issue with that is that the default value does not look great in the docs, but I think that's the price to pay for backwards compatibility in this PR, unless we wanted to just fall back to `ccc.nvim`s own defaults.

modules/plugins/utility/ccc/ccc.nix
jfeo: I decided to change this to `attrsOf anything` as well. I think that's what you meant as well? At least I don't see how a list would work here.

modules/plugins/utility/ccc/ccc.nix
jfeo: Thanks, I wasn't aware of the command role. I used it to refer to `:help ccc` in some places as well.

modules/plugins/utility/ccc/ccc.nix
jfeo: I actually did change it to a list in the latest commit on the branch. I can revert if you don't like it, but to me `listOf luaInline` seems to work alright for.

modules/plugins/utility/ccc/ccc.nix
jfeo: Changed this in the current commit :+1:

jfeo:
@NotAShelf Happy new year! 🎆 I have rebased this branch, and moved the release notes to the version 0.9 document.

Should I resolve the threads, or is that the reviewer who does that?

modules/plugins/utility/ccc/ccc.nix
NotAShelf: Either I was looking at the wrong field in ccc's help text, or I confused myself with something else. Yes, `attrsOf anything` seems good.

NotAShelf:
Happy new year! I've closed the review comments as you seem to have addressed them adequately. Looking at this a little late as I've spent most of today working on a different project, but I'm free (and with no strings attached!) tomorrow, so I'll provide another review. Though, all should be well.

If I do end up forgetting, please feel free to ping me anytime :)

jfeo:
A friendly ping @NotAShelf  :smile:

modules/plugins/utility/ccc/ccc.nix
```diff
@@ -1,10 +1,154 @@
 {lib, ...}: let
-  inherit (lib.options) mkEnableOption;
+  inherit (lib.options) mkOption mkEnableOption;
+  inherit
+    (lib.types)
+    anything
+    attrsOf
+    listOf
+    enum
+    ;
```
NotAShelf: I prefer not to waste time with style nits, but this is a reocurring problem that I'd rather not deal with in the long term. In most *sane* formatters, this remains as a single line. If formatted with nixfmt, Alejandra cannot fix it. It's preferrable that nixfmt is not used with this codebase.

NotAShelf (changes requested):
One tiny nit, LGTM otherwise.

NotAShelf:
Sorry for the delay, power was out most of yesterday. There is one tiny blocker and I *really* prefer not to nit those kinds of things, but it also doesn't get fixed unless someone stumbles across by luck. Can be merged after the formatting is fixed.

modules/plugins/utility/ccc/ccc.nix
jfeo: Apologies for not properly formatting, it should be fixed now.

I need to look into my own nvf config, since it seems that `nil_ls` uses `nixfmt` by default :)

NotAShelf (approved):

jfeo:
@NotAShelf Should I resolve the comments?

NotAShelf:
They... shouldn't block auto-merge but go ahead.

Soliprem:
marked them as resolved

jfeo:
Thanks! Hopefully my next contribution will require less feedback :smile:

PR #1293, opened 2025-12-16
meta: fix CONTRIBUTING.md 404 with absolute links

jfeo:
While looking into how to create a PR for a different issue, I noticed that reading the contribution guide on the repository home page by selecting the "Contributing" tab above the readme area, the relative links in `CONTRIBUTING.md` were not resolved as being relative to the `.github` directory, but rather relative to the repository root.

This PR just changes those links to from relative to absolute paths within the repository. It appears that GitHub properly maps absolute paths to be relative to the repository URL. I guess it could have an impact on mirrors or forks hosted elsewhere than GitHub, but since this relates to a file under the `.github` folder it seems reasonable to focus on it working well on GitHub :smile: 

I tested it both using the "Contributing" tab, and opening `.github/CONTRIBUTING.md` directly, and this way works for both.

First PR from me, it seems most of the contribution guidelines naturally target actual functionality, so I didn't check too many boxes below. Let me know if I missed anything!

Below is a screenshot of the issue:

nvf-pr-contributing-404

## Sanity Checking

[editorconfig]: https://editorconfig.org
[changelog]: https://github.com/NotAShelf/nvf/tree/main/docs/release-notes
[hacking nvf]: https://notashelf.github.io/nvf/index.xhtml#sec-guidelines

- [ ] I have updated the [changelog] as per my changes
- [x] I have tested, and self-reviewed my code
- [x] My changes fit guidelines found in [hacking nvf]
- Style and consistency
  - [ ] I ran **Alejandra** to format my code (`nix fmt`)
  - [ ] My code conforms to the [editorconfig] configuration of the project
  - [ ] My changes are consistent with the rest of the codebase
- If new changes are particularly complex:
  - [ ] My code includes comments in particularly complex areas
  - [ ] I have added a section in the manual
  - [ ] _(For breaking changes)_ I have included a migration guide
- Package(s) built:
  - [ ] `.#nix` _(default package)_
  - [ ] `.#maximal`
  - [ ] `.#docs-html` _(manual, must build)_
  - [ ] `.#docs-linkcheck` _(optional, please build if adding links)_
- Tested on platform(s)
  - [ ] `x86_64-linux`
  - [ ] `aarch64-linux`
  - [ ] `x86_64-darwin`
  - [ ] `aarch64-darwin`

---

Add a :+1: [reaction] to [pull requests you find important].

[reaction]: https://github.blog/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/
[...]

PR #1299, opened 2025-12-18
docs/manual: move v0.9 entries to the correct file; format

NotAShelf:
This is more or less a followup to #1293, with updates to the general "hacking" guidelines for areas that I felt were lacking or outdated. This should make the default contributor experience a bit smoother, with more concrete examples.

Fixes the incorrect changelog entry, and adds the 0.9 changelog document.

### DIFF
--- a/docs/manual/release-notes/rl-0.9.md
+++ b/docs/manual/release-notes/rl-0.9.md
@@ -2,8 +2,40 @@
 
 ## Changelog {#sec-release-0-9-changelog}
 
+## Breaking changes
+
+- Nixpkgs merged a full and incompatible rewrite of vimPlugins.nvim-treesitter.
+  The changes affected how grammars are built and it caused issues when neovim
+  attempted to load languages and could not find files in expected locations.
+
+## Changelog {#sec-release-0-9-changelog}
+
+[suimong](https://github.com/suimong):
+
+- Fix `vim.tabline.nvimBufferline` where `setupOpts.options.hover` requires
+  `vim.opt.mousemoveevent` to be set.
+
+[thamenato](https://github.com/thamenato):
+
+- Attempt to adapt nvim-treesitter to (breaking) Nixpkgs changes. Some treesitte grammars
+  were changed to prefer `grammarPlugins` over `builtGrammars`.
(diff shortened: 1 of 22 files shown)


docs/manual/release-notes/rl-0.9.md
```diff
@@ -0,0 +1,9 @@
+# Release 0.5 {#sec-release-0-9}
```
horriblename:
```suggestion
# Release 0.9 {#sec-release-0-9}
```

erm I'm akshually 9 comes after 8 :nerd_face:

NotAShelf: No way

Soliprem (approved):

flake.nix
```diff
@@ -81,7 +87,16 @@
...
-            alejandra --check ${self} < /dev/null | tee $out
+            alejandra --check ${self} < /dev/null
+            touch $out
+          '';
+
+          # Check if Markdown sources are properly formatted
+          # This can be initiated with `nix build .#checks.<system>.md-fmt`
+          # or with `nix flake check`
+          md-fmt = pkgs.runCommand "md-fmt-check" {nativeBuildInputs = [pkgs.deno];} ''
+            deno fmt --check ${self} --ext md
```
alfarelcynthesis, line 98: In case you aren't already aware, this line doesn't actually limit itself to markdown files, for some reason.
This works: `deno fmt --check ${self}/docs/manual ${self}/README.md`
but that's a workaround.

NotAShelf: I was not aware, and I'm wildly confused by how it could happen. Maybe something about the ordering...

alfarelcynthesis: `deno fmt --ext md --check ${self}` also doesn't respect the extension, so it might be a quirk of `--check`? I can't find any related issues, though.

NotAShelf: It does seem to be a quirk. Maybe it's intended.

I'll bring it up with upstream at a later date, but for now the above `fd` trick works for checking as well. Albeit a bit *loudly*. No matter, we can just silence the check.

PR #1304, opened 2025-12-22
Per Language Indent Config

mewoocat:
## Sanity Checking

- [ ] I have tested, and self-reviewed my code
- [ ] My changes fit guidelines found in [hacking nvf]
- Style and consistency
  - [ ] I ran **Alejandra** to format my code (`nix fmt`)
  - [ ] My code conforms to the [editorconfig] configuration of the project
  - [ ] My changes are consistent with the rest of the codebase
- Package(s) built:
  - [ ] `.#maximal`
- Tested on platform(s)
  - [ ] `x86_64-linux`
  - [ ] `aarch64-linux`
  - [ ] `x86_64-darwin`
  - [ ] `aarch64-darwin`

---

### DIFF
--- a/configuration.nix
+++ b/configuration.nix
@@ -51,9 +51,9 @@ isMaximal: {
       # Languages that will be supported in default and maximal configurations.
       nix = {
         enable = true;
-        tabstop = 20;
-        softtabstop = 20;
-        shiftwidth = 20;
+        tabstop = 8;
+        softtabstop = 8;
+        shiftwidth = 8;
       };
       markdown.enable = true;
 
--- a/modules/plugins/languages/nix.nix
+++ b/modules/plugins/languages/nix.nix
@@ -110,7 +110,7 @@ in {
 
     tabstop = mkOption {
       description = "Sets the tabstop size in spaces for .nix files";
-      type = nullOr null;
+      type = nullOr int;
       default = null;
     };
 


lib/languages.nix
```diff
@@ -77,4 +78,26 @@ in {
...
+    vim.autocmds = [
+      {
+        desc = "Sets indent for nix files";
+        event = ["BufEnter"];
+        pattern = [ "*.${language}" ];
+        callback = mkLuaInline ''
+          function()
+            vim.opt.tabstop = ${toString indentSize}
+            vim.opt.softtabstop = ${toString indentSize}
+            vim.opt.shiftwidth = ${toString indentSize}
```
horriblename: - `vim.opt` should only be used when appending/removing entries from list-type options, it may be deprecated in the future
- filetype specific options should use `vim.opt_local` or `vim.bo` or `vim.wo` depending on context

lib/languages.nix
```diff
@@ -77,4 +78,26 @@ in {
...
+
+  # maybe put generic function to set indent for provided langs
+  # Then we can just call it with the lib arg
+
+  setLanguageIndent = {language, indentSize}: {
+    vim.autocmds = [
+      {
+        desc = "Sets indent for nix files";
+        event = ["BufEnter"];
+        pattern = [ "*.${language}" ];
```
horriblename: detecting FileType is harder than just matching the file extension, e.g. flake.lock is a json file but does not match `*.json`. A file with no extension in the name but with content starting with a shebang `#!/bin/python` is detected as a python file, but again does not match `*.py`. BufEnter does not cover all edge cases and makes adding more edge cases even harder. Use the `FileType` event if you need to match file types

mewoocat:
@horriblename 

> I don't think these kinds of options are needed, vim typically provides a good default for most filetypes so it feels weird providing the same default for most filetypes. If you feel like vim's default is not good, consider making a PR upstream.

oh I wasn't aware vim had any per filetype defaults, I'm not really finding much info on this.  I had always thought the tab configuration was static.

> If the goal is a more convenient per-filetype option API, there are better ways to do so, e.g. mimicking the [`vim.options`](https://nvf.notashelf.dev/options.html#option-vim-options) API, but for each filetype.

I'm not sure I follow.  Are you suggesting something like `vim.options.<filetype>.indentSize` as opposed to what I have now `vim.languages.<lang>.indentSize`?

lib/languages.nix
mewoocat: switched to vim.bo

lib/languages.nix
mewoocat: switched to filetype event

horriblename:
> I'm not sure I follow. Are you suggesting something like vim.options.<filetype>.indentSize as opposed to what I have now vim.languages.<lang>.indentSize?

something like `vim.perFileTypeOptions.<lang>.shiftwidth`

though I'd ask for opinions from @NotAShelf and @Soliprem first

NotAShelf:
> > I'm not sure I follow. Are you suggesting something like vim.options..indentSize as opposed to what I have now vim.languages..indentSize?
> 
> something like `vim.perFileTypeOptions.<lang>.shiftwidth`
> 
> though I'd ask for opinions from @NotAShelf and @Soliprem first

I'm not sure honestly. Adding yet another namespace under `vim` just to handle options per-file seems... ugly. What I had mind recently was an abstraction to write what is essentially Neovim's ftplugin files in the runtime directory, and perhaps we could abstract around that. That said, I'm pretty happy to keep it under `vim.languages` since that *is* our current language abstraction. Though, there needs to be a toggle *which is off by default*. We do not want to set defaults for people. 

P.S.: rather than this:

```lua
          function()
            vim.bo.tabstop = ${toString indentSize}
            vim.bo.softtabstop = ${toString indentSize}
            vim.bo.shiftwidth = ${toString indentSize}
          end
```

we'll want each option to be customizable separately.

P.P.S: Allow nulling the values to avoid setting the option altogether. If `cfg.tabstop` is `null`, it should not be set and so on.

mewoocat:
@NotAShelf what do you think about these changes?

snoweuph:
I personally see little reason for this, as `.editorconfig` exists, neovim will respect it: <https://neovim.io/doc/user/plugins/#_builtin-plugin%3a-editorconfig>

Yes, this is per-project config, not global config, but at the point where you share your projects with others, you prolly will need to add one anyways

mewoocat:
> I personally see little reason for this, as `.editorconfig` exists, neovim will respect it: https://neovim.io/doc/user/plugins/#_builtin-plugin%3a-editorconfig
> 
> Yes, this is per-project config, not global config, but at the point where you share your projects with others, you prolly will need to add one anyways

I agree, I originally was going into this without much knowledge of .editorconfig.  But now Im more aware, I don't think my changes here are worth it.

NotAShelf:
I'm in favor of keeping this change. `.editorconfig` is not *properly* standardized, and I don't see a reason to enforce it just because it exists.

mewoocat:
> I'm in favor of keeping this change. `.editorconfig` is not _properly_ standardized, and I don't see a reason to enforce it just because it exists.

Wdym not properly standardized?

NotAShelf:
Not all LSPs or formatters support it. An autocmd for setting the indent, however, will always work.

mewoocat:
Well if you're cool with current implementation here I'll try to publish the PR sometime in the near future.

snoweuph:
> Not all LSPs or formatters support it.

some don't support it, because the language standardized the indentations, like go and make.

But yes, there are LSPs out there, that don't support the protocol, or act as IF (im talking about you tombi the AI slop)

Issue #1305, opened 2025-12-23
Deprecating `setupOpts` options


imnotpoz:
due to `setupOpts` not being an attrset trying to deprecate an option using `mkRemovedOptionModule` documented inside them leads to `error: evaluation aborted with the following error message: 'cannot find attribute 'vim.*''`

what's the way to fix this, do we make our own `mkRemovedOptionModule` or some other solution?

horriblename:
~~what's the exact situation? off the top of my head `mkOption { visible = false; apply = lib.warn "option is deprecated";} ` should work fine~~ not it doesn't

imnotpoz:
https://github.com/imnotpoz/nvf/commit/04931eccde9a8215a15ee9bc59645216050f48ff#diff-61e35c9752d03834a1705905b61247d2bdd07e10eae55e70f51cdcaeceae9c9aR20-R30

imnotpoz:
> what's the exact situation? off the top of my head `mkOption { visible = false; apply = lib.warn "option is deprecated";} ` should work fine

I'll try that

imnotpoz:
this works thank you

PR #1306, opened 2025-12-23
fidget-nvim setupOpts update

imnotpoz:
found these in `:checkhealth`

## Sanity Checking

- Style and consistency
- Package(s) built:
  - [ ] `.#maximal`
- Tested on platform(s)
  - [x] `x86_64-linux`
  - [x] `aarch64-linux`
  - [ ] `x86_64-darwin`
  - [ ] `aarch64-darwin`

---

### DIFF
--- a/modules/plugins/visuals/fidget-nvim/fidget.nix
+++ b/modules/plugins/visuals/fidget-nvim/fidget.nix
@@ -53,6 +53,24 @@ in {
             end
           '';
         };
+        clear_on_detach = mkOption {
+          type = nullOr luaInline;
+          default = mkLuaInline ''
+            function(client_id)
+              local client = vim.lsp.get_client_by_id(client_id)
+              return client and client.name or nil
+            end
+          '';
+          defaultText = literalExpression ''
+            default = mkLuaInline '''
+              function(client_id)
+                local client = vim.lsp.get_client_by_id(client_id)
+                return client and client.name or nil
+              end
+            ''';
+          '';
+          description = "Clear notification group when LSP server detaches";
+        };
         ignore = mkOption {
           description = "Ignore LSP servers by name";
           type = listOf str;
@@ -417,6 +435,20 @@ in {
             type = bool;
             default = true;
           };
+          align = mkOption {
+            type = enum ["message" "annote"];
+            default = "message";
+            description = "Indent messages longer than a single line";
+          };
+          reflow = mkOption {
+            type = enum ["hard" "hyphenate" "ellipsis" "false"];
+            default = "false";
+            description = ''
+              Reflow (wrap) messages wider than notification window
+
+              The various options determine how wrapping is handled mid-word.
+            '';
+          };
           icon_separator = mkOption {
             description = "Separator between group name and icon";
             type = str;
@@ -431,6 +463,16 @@ in {
             description = "Highlight group used for group separator";
             type = str;
             default = "Comment";
+          };
+          line_margin = mkOption {
+            type = int;
+            default = 1;
+            description = ''
+              Spaces to pad both sides of each non-empty line
+
+              Useful for adding a visual gap between notification text
+              and any buffer it may overlap with.
+            '';
           };
           render_message = mkOption {
             description = "How to render notification messages";
@@ -461,6 +503,15 @@ in {
               if config.vim.ui.borders.enable
               then config.vim.ui.borders.globalStyle
               else "none";
+          };
+          border_hl = mkOption {
+            type = str;
+            default = "";
+            description = ''
+              Highlight group for notification window border
+
+              Set to empty string to keep your theme's default `FloatBorder` highlight.
+            '';
           };
           zindex = mkOption {
             description = "Stacking priority of the notification window";
@@ -497,25 +548,15 @@ in {
             type = enum ["editor" "win"];
             default = "editor";
           };
-        };
-      };
-
-      integration = {
-        nvim-tree = {
-          enable = mkOption {
-            description = "Integrate with nvim-tree/nvim-tree.lua (if enabled)";
-            type = bool;
-            default =
-              if config.vim.filetree.nvimTree.enable
-              then true
-              else false;
+          tabstop = mkOption {
+            type = int;
+            default = 8;
+            description = "Width of each tab character in the notification window";
           };
-        };
-        xcodebuild-nvim = {
-          enable = mkOption {
-            description = "Integrate with wojciech-kulik/xcodebuild.nvim (if enabled)";
-            type = bool;
-            default = true;
+          avoid = mkOption {
+            type = listOf str;
+            default = [];
+            description = "Filetypes the notification window should avoid";
           };
         };
       };
@@ -545,6 +586,46 @@ in {
           '';
         };
       };
+
+      # removed, see below
+      integration = {
+        nvim-tree.enable = mkOption {
+          default = null;
+          visible = false;
+        };
+        xcodebuild-nvim.enable = mkOption {
+          default = null;
+          visible = false;
+        };
+      };
     };
+  };
+
+  # this can't be done better, I tried
+  # mostly mostly caused by the deprecated options being inside a submodule
+  # try improving this if you don't care about your sanity
+  # ~ poz
+  config = {
+    assertions = let
+      inherit (config.vim.visuals.fidget-nvim.setupOpts) integration;
+    in [
+      {
+        assertion = integration.nvim-tree.enable == null;
+        message = ''
+          Option `vim.visuals.fidget-nvim.setupOpts.integration.nvim-tree.enable`
+          has been deprecated upstream. Use
+          `vim.visuals.fidget-nvim.setupOpts.notification.window.avoid = ["NvimTree"]` instead.
+          This is already set if `vim.filetree.nvimTree.enable == true`.
+        '';
+      }
+      {
+        assertion = integration.xcodebuild-nvim.enable == null;
+        message = ''
+          Option `vim.visuals.fidget-nvim.setupOpts.integration.xcodebuild-nvim.enable`
+          has been deprecated upstream. Use
+          `vim.visuals.fidget-nvim.setupOpts.notification.window.avoid = ["TestExplorer"]` instead.
+        '';
+      }
+    ];
   };
 }
--- a/modules/plugins/visuals/fidget-nvim/fidget.nix
+++ b/modules/plugins/visuals/fidget-nvim/fidget.nix
@@ -25,33 +25,33 @@ in {
     setupOpts = mkPluginSetupOption "Fidget" {
       progress = {
         poll_rate = mkOption {
-          description = "How frequently to poll for LSP progress messages";
           type = int;
           default = 0;
+          description = "How frequently to poll for LSP progress messages";
         };
         suppress_on_insert = mkOption {
-          description = "Suppress new messages when in insert mode";
           type = bool;
           default = false;
+          description = "Suppress new messages when in insert mode";
         };
         ignore_done_already = mkOption {
-          description = "Ignore new tasks that are already done";
           type = bool;
           default = false;
+          description = "Ignore new tasks that are already done";
         };
         ignore_empty_message = mkOption {
-          description = "Ignore new tasks with empty messages";
           type = bool;
           default = false;
+          description = "Ignore new tasks with empty messages";
         };
         notification_group = mkOption {
-          description = "How to get a progress message's notification group key";
           type = luaInline;
           default = mkLuaInline ''
             function(msg)
               return msg.lsp_client.name
             end
           '';
+          description = "How to get a progress message's notification group key";
         };
         clear_on_detach = mkOption {
           type = nullOr luaInline;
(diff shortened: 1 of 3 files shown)


modules/plugins/visuals/fidget-nvim/fidget.nix
```diff
@@ -505,17 +559,27 @@ in {
...
             description = "Integrate with wojciech-kulik/xcodebuild.nvim (if enabled)";
             type = bool;
-            default = true;
+            default = false;
+            visible = false;
+            apply = warn ''
+              Option `vim.visuals.fidget-nvim.setupOpts.integration.xcodebuild-nvim.enable`
+              has been deprecated upstream. Use
+              `vim.visuals.fidget-nvim.setupOpts.notification.window.avoid = ["TestExplorer"]` instead.
+            '';
```
NotAShelf: This is not the correct way to warn on an option. You'll want to use `mkRemovedOptionModule` as the rest of the codebase. Apply does work, but it's a really fragile way of doing this. Also applies to the `vim.visuals.fidget-nvim.setupOpts.integration.nvim-tree.enable` warning above.

modules/plugins/visuals/fidget-nvim/fidget.nix
```diff
@@ -497,6 +541,16 @@ in {
...
           };
+          tabstop = mkOption {
+            description = "Width of each tab character in the notification window";
+            type = int;
+            default = 8;
+          };
+          avoid = mkOption {
+            description = "Filetypes the notification window should avoid";
+            type = listOf str;
+            default = [];
```
NotAShelf, line 558: Please order `mkOption` args as 

1. type
2. default
3. (optional) example
4. Description

Also fwiw the `setupOpts` block is freeform so new options are not strictly necessary. Though if you *do* want to type-check those or for those to show up on documentation, it's fine.

NotAShelf (changes requested):
A few comments on formatting and introduction of new options. I'd appreciate it if you could follow the "convention" for `mkOption`.

Also, this requires a changelog entry after v0.9. LGTM otherwise. Though CC @horriblename for the HLS changes.

modules/plugins/visuals/fidget-nvim/fidget.nix
imnotpoz: unfortunately not possible due to `mkPluginSetupOption`, see https://github.com/NotAShelf/nvf/issues/1305

modules/plugins/visuals/fidget-nvim/fidget.nix
imnotpoz: every single option in this module is ordered `description -> type -> default`, should I change all of them to this ordering?

imnotpoz:
> Also, this requires a changelog entry after v0.9. LGTM otherwise. Though CC @horriblename for the HLS changes.

changelog for both commits? I can add it once the comments above are resolved

horriblename (changes requested):
I'm pretty this will trigger the deprecation warning regardless of whether the user has set this option

I _think_ removing the default value works? (I don't remember if it allows not giving a default without one)

modules/plugins/languages/haskell.nix
```diff
@@ -20,10 +20,11 @@
 
   defaultServers = ["hls"];
   servers = {
+    # these are a haskell-tools wrapper over the actual lsp options `:help haskell-tools.lsp.ClientOpts`
+    # see https://github.com/mrcjkb/haskell-tools.nvim/blob/v6.2.0/lua/haskell-tools/lsp/init.lua#L131-L173
+    # for the real ones
     hls = {
-      enable = false;
```
horriblename, line 24: from my understanding, we're relying on haskell-tools to start the lsp, so we should set this to false? (enable = true, which is the default, calls vim.lsp.enable("server_name")

horriblename (changes requested):

modules/plugins/languages/haskell.nix
imnotpoz: `enable` is one of the fields not recognized by haskell-tools, as per `:checkhealth`:

```
Checking config ~
- ✅ OK No errors found in config.
- ⚠️ WARNING unrecognized configs in vim.g.haskell_tools: { "hls.enable", "hls.root_dir", "hls.filetypes" }
```

in that commect I left there, `lsp_start_opts` use *some* fields from `hls_opts` and `hls_settings`, which are the ones defined by us - there's no checks on `enable`

imnotpoz:
repushed with a rebase on current main

imnotpoz:
[this](https://github.com/NotAShelf/nvf/actions/runs/22143345368/job/64013344890?pr=1306#step:5:108) doesn't seem to be caused by my PR I think

modules/plugins/languages/haskell.nix
imnotpoz: just tested this on this branch, there's no warnings without `enable` and the LSP still works properly

horriblename: hmm in that case we should filter out `enable` before toLuaObject'ing it. I can do that in a separate PR, or you can do that in this one

imnotpoz: do you mean adding that `enable` back, setting it to `true` so that it stays consistent with the rest, but filtering that `enable` when passed to `toLuaObject`?

modules/plugins/visuals/fidget-nvim/fidget.nix
imnotpoz: resolved in the latest commit

modules/plugins/languages/haskell.nix
horriblename: `enable = false` should be kept, and enable should be filtered out before being passed to toLuaObject [here](https://github.com/NotAShelf/nvf/blob/7dc5b7980026136c11b4c704508289876e55d8ee/modules/neovim/init/lsp.nix?plain=1#L23). keep in mind that, the enable field is only used by nvf to call `vim.lsp.enable('lspname')`, and has no effect when passed to `vim.lsp.config()`

horriblename: ok my bad, there was another effort in fixing this already #1191 

I'll read through the discussions in that again later, I kinda forgot about it.

I think it's best to just leave haskell-tools alone for now, I'll merge in the other changes

imnotpoz: repushed and removed the haskell commit

modules/plugins/visuals/fidget-nvim/fidget.nix
horriblename: sorry I lied xd, this throws a warning even if the option isn't set at all. We'll have to figure something out

modules/plugins/visuals/fidget-nvim/fidget.nix
```diff
@@ -25,69 +26,78 @@ in {
...
           '';
+          description = "How to get a progress message's notification group key";
+        };
+        clear_on_detach = mkOption {
+          type = nullOr luaInline;
+          default = mkLuaInline ''
+            function(client_id)
+              local client = vim.lsp.get_client_by_id(client_id)
+              return client and client.name or nil
+            end
```
NotAShelf, line 62: We can also add a `defaultText` after this:

```
defaultText = literalExpression '''
          default = mkLuaInline ''
            function(client_id)
              local client = vim.lsp.get_client_by_id(client_id)
              return client and client.name or nil
            end
          '';
''';
```

NotAShelf (dismissed):
Everything LGTM. One comment, but it's not a blocker.

modules/plugins/visuals/fidget-nvim/fidget.nix
imnotpoz: sure

imnotpoz: you mean something like this? I feel like I'm misunderstanding

```nix
        clear_on_detach = mkOption {
          description = "Clear notification group when LSP server detaches";
          type = nullOr luaInline;
          default = mkLuaInline ''
            function(client_id)
              local client = vim.lsp.get_client_by_id(client_id)
              return client and client.name or nil
            end
          '';
          defaultText = literalExpression ''
            default = mkLuaInline '''
              function(client_id)
                local client = vim.lsp.get_client_by_id(client_id)
                return client and client.name or nil
              end
            ''';
          '';
        };
```

imnotpoz: or whatever way you escape `''` inside of another `''`

imnotpoz: @NotAShelf repushed with what I think your intent was

NotAShelf: I managed to fuck that up, yes that was it thanks.

NotAShelf (dismissed):

imnotpoz:
formatting check fails because it can't build deno xddddd

NotAShelf:
Blame Nixpkgs. Can be merged with @horriblename's approval.

imnotpoz:
> Blame Nixpkgs. Can be merged with @horriblename's approval.

yeah I reported that [issue](https://github.com/NixOS/nixpkgs/issues/511900)

snoweuph (commented):
LGTM, but maybe add atleast a sentence to the changelog, as there seems to be some fixes besides just doc changes.

imnotpoz:
> LGTM, but maybe add atleast a sentence to the changelog, as there seems to be some fixes besides just doc changes.

done, should be good now

snoweuph:
@imnotpoz I suggest you rebase your branch (or merge main into it), and close all the threads as completed, that are completed and than ping `@`horriblename, as his review is blocking merge RN.

image

Also Ive just noticed your CI is failing, becuse the flake is invalid

imnotpoz:
@horriblename review? :pleading_face:

> Also Ive just noticed your CI is failing, becuse the flake is invalid

this is caused by nixpkgs, as stated [above](https://github.com/NotAShelf/nvf/pull/1306#issuecomment-4353173075)

modules/plugins/visuals/fidget-nvim/fidget.nix
horriblename: does this actually work as intended?

snoweuph:
I'm not talking about deno though.
im talking bout `nix flake check` faling. look at the CI.

```sh
error: The option `vim.visuals.fidget-nvim.setupOpts.progress.clear_on_detach' was accessed but has no value defined. Try setting the option.
```

imnotpoz:
> I'm not talking about deno though. im talking bout `nix flake check` faling. look at the CI.
> 
> ```shell
> error: The option `vim.visuals.fidget-nvim.setupOpts.progress.clear_on_detach' was accessed but has no value defined. Try setting the option.
> ```

sorry, missed that somehow

no errors in `nix flake check` on my machine now, although I'm getting the nvim tree integration warning 4 times - can't find where that can happen

modules/plugins/visuals/fidget-nvim/fidget.nix
imnotpoz: sorry, finally got back around to this PR - even if not set this warning does pop up quite a lot (the rest seems to work though) and `mkRemovedOptionModule` doesn't work at all so we do have to figure something else out

snoweuph:
no nicks from me then.

@horriblename could you do a final review?

image

imnotpoz:
> no nicks from me then.
> 
> @horriblename could you do a final review?
> image

there's still one issue to fix, see the only unresolved discussion above

imnotpoz:
please check the 2 latest commits, this still doesn't work due to an infinite recursion that I don't know what to do with

what I'm doing here right now seems exactly the same when comparing to the linked snippet but as it's past 3 in the morning I might be wrong

either way, considering all the other options don't work and the fact that this is taken from nixpkgs, I believe this is the right way to go with this

while debugging I found that removing line 15 in `modules/plugins/visuals/fidget-nvim/config.nix` (which is the only usage of `vim.visuals.fidget-nvim.setupOpts` in all of nvf aside from these assertions) will make the infinite recursion disappear and the whole thing builds and I believe this is the best lead I have so far

imnotpoz:
the latest commit, although ugly and breaks nvf conventions seems to have fixed this, no infinite recursion on my machine and builds successfully - waiting for some opinions on this and whether we want to go forward with it

imnotpoz:
great, now the docs hit an infinite recursion

will fix this later I am tired

snoweuph:
> great, now the docs hit an infinite recursion
> 
> will fix this later I am tired

Good might, sleep well

imnotpoz:
`nix flake check` fully passes on my machine!!!!!

imnotpoz:
@NotAShelf @horriblename please review the current deprecation method

imnotpoz:
I am so sorry for the ping, turns out setupOpts are being set inside of setupOpts

imnotpoz:
pushed some bullshit that doesn't work either and still gives infinite recursion

to use the mention of raf and horriblename, I'll ask you if it'd be ok to set both these integration options to `false` by default, make them not visible and only throw if they're set to true? this would simplify pretty much everything and won't require any of this bullshit

horriblename:
> I'll ask you if it'd be ok to set both these integration options to false by default, make them not visible and only throw if they're set to true? this would simplify pretty much everything and won't require any of this bullshit

that's fine, except use null so nothing ends up in init.lua (types will need to be adjusted as well)

horriblename:
~~I have another idea, but using null and checking for non-default is honestly good enough I think~~

~~just as a note to myself in case null is somehow inadequate in the future: I'm pretty sure some incantation of `coercedTo (warn "...") (/* nullOr? */ TYPE)` can work~~

no it doesn't I'm dumb (still doesn't do a rename, not any better than just asserting null)

imnotpoz:
pushed something that seems to work, will cleanup commit history some time after I wake up

please do review the actual changes, there should be no more outside of the history being cleaned up

modules/plugins/visuals/fidget-nvim/fidget.nix
```diff
@@ -409,142 +427,212 @@ in {
...
+          This is already set if `vim.filetree.nvimTree.enable == true`.
+        ''
+      else if integration.xcodebuild-nvim.enable != null
+      then
+        throw ''
+          Option `vim.visuals.fidget-nvim.setupOpts.integration.xcodebuild-nvim.enable`
+          has been deprecated upstream. Use
+          `vim.visuals.fidget-nvim.setupOpts.notification.window.avoid = ["TestExplorer"]` instead.
+        ''
+      else null;
```
horriblename: use the warnings/assertions option instead:

```
conifg.assertions = [
  { condition = cfg... == null;
    message = ""
 } 
] ;
```

horriblename (changes requested):

imnotpoz:
repushed to clean up history, just 2 commits now :3 all works and ready for review!!!

NotAShelf (approved):

horriblename (approved):
thanks!

PR #1316, opened 2026-01-06
various: continue migrating to new nvim-treesitter API

NotAShelf:
Might address the breaking Treesitter syntax highlighting. Seems to work for Nix, and I imagine it works for the rest of them. Unfortunately I cannot test other languages. Nor do I want to.

### DIFF
--- a/lib/types/languages.nix
+++ b/lib/types/languages.nix
@@ -30,7 +30,7 @@
 
   mkGrammarOption = pkgs: grammar:
     mkPackageOption pkgs ["${grammar} treesitter"] {
-      default = ["vimPlugins" "nvim-treesitter" "grammarPlugins" grammar];
+      default = ["vimPlugins" "nvim-treesitter" "builtGrammars" grammar];
     };
 in {
   inherit diagnostics diagnosticSubmodule mkGrammarOption;
--- a/modules/plugins/languages/python.nix
+++ b/modules/plugins/languages/python.nix
@@ -228,7 +228,7 @@ in {
       package = mkOption {
         description = "Python treesitter grammar to use";
         type = package;
-        default = pkgs.vimPlugins.nvim-treesitter.grammarPlugins.python;
+        default = pkgs.vimPlugins.nvim-treesitter.builtGrammars.python;
       };
     };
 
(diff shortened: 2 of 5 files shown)


ethanh20009:
Tested, my findings:

Enabling text objects now doesn't through any errors.
but can't use existing setup for auto folds, toggling etc. No ts folds work at all.

I get TS highlights for my languages, but they got much worse. It's possible the names have changed for tokens and therefore the color scheme no longer matches.
However, in a lot of cases, I get different :Inspect values returned for the same parsers. Mostly with the new treesitter, I get pretty much no hits on :Inspect with angular parser on html templates.
:InspectTree seems to work, and get the basic structure.
My thinking is it might be defaulting to html parser somehow, because it identifies tags, but not things like @if in html.
but I run `:lua print(vim.treesitter.get_parser():lang())` and I get `angular` in both old and new treesitter :shrug:

ethanh20009:
https://github.com/user-attachments/assets/7118beee-a800-403e-8aeb-763bc630d0f9

Almost figured it out...
By toggling vim.treesitter.start / stop, I can get each half of the highlighting I want.
But not both at the same time like it does before the new treesitter breaking changes.

Also, it doesn't look like vim.treesitter.start is being called automatically. 

I can get it to work for angular html if I use

```nix
treesitter.grammars = with pkgs.vimPlugins.nvim-treesitter.parsers;
  [
    nix
    lua
    angular
    html
    json
    javascript
    rust
    typescript
  ]
  ++ (with pkgs.vimPlugins.nvim-treesitter.queries; [
    nix
    lua
    angular
    html
    json
    javascript
    rust
    typescript
  ]);

luaConfigRC.treesitter = ''
  vim.api.nvim_create_autocmd('FileType', {
    pattern = { 'nix', 'lua', 'angular', 'typescript', 'ts', 'javascript', 'html', 'json', 'rust', 'js' },
    callback = function() vim.treesitter.start() end,
  })
'';

```

Working for most (html, lua, nix, rust, json) but it doesn't work for typescript...

NotAShelf:
#1324 was merged earlier, I'll rebase this shortly and see if that resolves most of the issues.

NotAShelf:
treesitter-textobjects updated in the main branch, and grammarPlugins was resolved by an earlier PR. No longer necessary.

PR #1321, opened 2026-01-07
terraform/hcl: improve options; fix terraform-cli error; add tofu-ls

ppenguin:
- Add tofu(ls) as free terraform alternative for terraform and hcl
- Harmonize with other format/LSP implementation (also with conform)
- Add/fix terraform formatter
- Add `nomad fmt` as alternative hcl formatter

Note: for (nomad) hcl tofu/terraform ls are ineffective (don't support e.g. jobspec schema) -> TODO: `nomad-lsp` (package and add, even though it seems abandoned it's probably better for nomad/hcl)

## Sanity Checking

- Style and consistency
- Package(s) built:
  - [x] `.#maximal`
- Tested on platform(s)
  - [x] `x86_64-linux`
  - [ ] `aarch64-linux`
  - [ ] `x86_64-darwin`
  - [ ] `aarch64-darwin`

---


modules/plugins/languages/hcl.nix
```diff
@@ -81,18 +95,24 @@ in {
...
+        lsp.servers =
+          mapListToAttrs (n: {
+            name = n;
+            value = servers.${n};
+          })
+          cfg.lsp.servers;
+        extraPackages =
+          (lib.optionals (elem "terraformls" cfg.lsp.servers) [pkgs.terraform])
+          ++ (lib.optionals (elem "tofuls" cfg.lsp.servers) [pkgs.opentofu]);
+      };
```
sjcobb2022, line 126: I don't think this is the best way to do things. It is not the responsibility of NVF to provide terraform or opentofu packages necessarily.

modules/plugins/languages/terraform.nix
```diff
@@ -4,57 +4,113 @@
...
-        cfg.lsp.servers;
+      vim = {
+        lsp.servers =
+          mapListToAttrs (n: {
+            name = n;
+            value = servers.${n};
+          })
+          cfg.lsp.servers;
+        extraPackages =
+          (lib.optionals (elem "terraformls" cfg.lsp.servers) [pkgs.terraform])
```
sjcobb2022: Same as above comment

sjcobb2022 (changes requested):
Looks good overall other than a few comments. 

In addition, if a user enables both HCL and terraform modules, the lsp configs may overwrite each other.

I know you have tried to combat this by using mkDefault, but I think a more reasonable solution may be to "tag" the LSPs with something like: terraformls-hcl, or opentofu-tf, this way we can have the configs fully separate. 

Alternatively, if the configs are identical other than the filetypes, then there is no need to use mkDefault as nix should merge the two arrays anyway.

modules/plugins/languages/hcl.nix
ppenguin: > I know you have tried to combat this by using mkDefault, but I think a more reasonable solution may be to "tag" the LSPs with something like: terraformls-hcl, or opentofu-tf, this way we can have the configs fully separate.

:+1:  done.

I added the cli tools because a while back (before the change) I got an error stating they were needed. But it seems to have been a red herring (coudn't reproduce that), so I removed as per your suggestion.

NotAShelf: If LSP needs a CLI package to *function*, it is usually provided via wrapping in Nixpkgs. If not, you can wrap the LSP using symlinkJoin---this is done in a few places in nvf, but not necessarily a clean solution.

I intend to add an option that decides whether nvf modules *install packages* or discover from PATH but I could not find a clean solution for that yet.

sjcobb2022:
Would definitely want @NotAShelf and @horriblename to take a look at this impl

modules/plugins/languages/hcl.nix
```diff
@@ -14,21 +14,33 @@
 
   cfg = config.vim.languages.hcl;
 
-  defaultServers = ["terraform-ls"];
+  defaultServers = ["tofuls-hcl"];
```
NotAShelf, line 17: How is compatibility between terraform-ls and tofuls-hcl? I've been aware of tofu's existence for a while, but I'd like to avoid breaking changes if that's a possibility.

modules/plugins/languages/hcl.nix
```diff
@@ -14,21 +14,33 @@
...
+      filetypes = ["hcl"];
+      root_markers = [".git"];
+    };
+    tofuls-hcl = {
+      enable = true;
+      cmd = [(getExe pkgs.tofu-ls) "serve"];
+      filetypes = ["hcl"];
       root_markers = [".terraform" ".git"];
     };
+    # TODO: package nomad-lsp and offer as an option here too
```
NotAShelf, line 31: You can package this in-house as we do for some plugins. Can be upstreamed to Nixpkgs at a later date.

modules/plugins/languages/terraform.nix
```diff
@@ -34,27 +54,57 @@ in {
 
     lsp = {
       enable = mkEnableOption "Terraform LSP support (terraform-ls)" // {default = config.vim.lsp.enable;};
-
       servers = mkOption {
         type = listOf (enum (attrNames servers));
         default = defaultServers;
-        description = "Terraform LSP server to use";
+        description = "Terraform LSP server to use (one or more of [${concatStringsSep " " (attrNames servers)}])";
```
NotAShelf: No need for this to be changed. `listOf enum` should already render available items.

modules/plugins/languages/terraform.nix
```diff
@@ -34,27 +54,57 @@ in {
...
+      };
+    };
+
+    format = {
+      enable = mkOption {
+        type = bool;
+        default = config.vim.languages.enableFormat;
+        description = "Enable Terraform formatting";
+      };
```
NotAShelf: Use `mkEnableOption` here as we do for other options.

NotAShelf (changes requested):

modules/plugins/languages/hcl.nix
ppenguin: According to [`terraform-ls` feature matrix](https://github.com/hashicorp/terraform-ls/blob/main/docs/features.md) and [`tofu-ls` feature matrix](https://github.com/opentofu/tofu-ls/blob/main/docs/features.md) they have feature parity, from my own experience there's also no difference.  `opentofu` and `terraform` behave also the same (also personal observation), but the latter has an unfree license. 

The reason I chose `tofu-ls` as default is purely because its "sibling" `opentofu` has a free license and would therefore be a less "invasive" default for a user (because it doesn't require to set `allowUnfree(Predicate)` on `nixpkgs`), even though `terraform-ls` itself has a free license.

modules/plugins/languages/terraform.nix
ppenguin: reverted

modules/plugins/languages/hcl.nix
ppenguin: Good idea, thanks. Will have to get back later to this though, since I'm a bit short on time now. (I expect some regressions for the build since the last commit on that repo is 5 years ago).

modules/plugins/languages/terraform.nix
ppenguin: done

ppenguin:
@NotAShelf BTW, there seems to be a regression on `main` (introduced sometime since 7 days ago), that breaks `nix build .#docs-html`:

```
error: Cannot build '/nix/store/ar4ndivdlidqp8glnbvag5fq2hvxsjdz-nvf-docs-html.drv'.
       Reason: builder failed with exit code 1.
       Output paths:
         /nix/store/3ixr2az6w2dqxzz745qn1zsxjz86gmmg-nvf-docs-html
...
       > Error:
       >    0: Configuration error: No config file provided. Use --config-file to specify a config file.
       >
       > Location:
       >    ndg/src/main.rs:106
```
This affected this issue as well since I did a rebase.

modules/plugins/languages/hcl.nix
NotAShelf: Makes sense, thanks for explaining. In any case I think its worth a mention under the breaking changes section even though it is not necessarily breaking; one sentence entry that bubbles up the change would be suitable with a note on compatibility.

NotAShelf:
Ah damn I forgot to test the config module before NDG's 2.6.0 refactor. I'll try to take a look today after work or tomorrow during the day.

modules/plugins/languages/hcl.nix
ppenguin: Oops, missed this point, now added. I think all requested changes are now complete?

modules/plugins/languages/terraform.nix
```diff
@@ -51,20 +71,47 @@ in {
         description = "Terraform LSP server to use";
       };
     };
+
+    format = {
+      enable = mkEnableOption "Enable Terraform formatting" // {default = config.vim.languages.enableFormat;};
+      type = mkOption {
+        type = deprecatedSingleOrListOf "vim.language.terraform.format.type" (enum (attrNames formats));
```
NotAShelf: Since this is a new addition, no need to be backwards compatible with the deprecated single-choice option. You can just make this a list.

modules/plugins/languages/terraform.nix
```diff
@@ -51,20 +71,47 @@ in {
         description = "Terraform LSP server to use";
       };
     };
+
+    format = {
+      enable = mkEnableOption "Enable Terraform formatting" // {default = config.vim.languages.enableFormat;};
```
NotAShelf: Needs `defaultText` as was introduced recently. This helps with eval times while building the documentation.

NotAShelf (changes requested):
Two quick notes to address. Looks good to me otherwise.

NotAShelf (approved):

NotAShelf:
Thank you for your patience, and for your efforts :)

PR #1345, opened 2026-01-15
npins: update to version 7

Gerg-L:
Gerg certified™ npins usage


NotAShelf:
One comment but I cba nitting rn, will fix manually.

PR #1365, opened 2026-01-27
various: internal refactor

NotAShelf:
This is an internal refactor that aims to be non-breaking as much as possible. There are two changes, with a few more on the way:

1. There is a new `vim.filetype` API that can be used to register custom filetypes using `vim.filetype.add()`. This is placed in the DAG hierarchy *before* `lazyConfigs` so it's possible to use the API to register custom filetypes that then get used to lazy-load a plugin on FT.
2. Moves `neovim/mappings` to `neovim/init/mappings.nix` since the neovim/init module is where I've decided to consolidate *core* Neovim options. This is no longer "init" only, so it'll be renamed in a later PR or on the master branch. Not a breaking change.

This PR has been inspired by, and aims to support, #1304 so I'll be adding `vim.ftplugin` and `vim.ftdetect` options as auxiliary module options that the per-language indents can be use with.

Last but not least, I'll go over existing core options and extend options or documentation where applicable. Also CC @horriblename and @Soliprem on something I am not quite sure about: I'd like modules to be able to create their documentation bits to create per-plugin pages, and while NDG *does* allow this, I am not sure what the API would look like. Should we create a `vim.docs` module that defines the pages? Is leaking documentation sources to user-space a waste of space?


Soliprem (dismissed):
LGTM.

On adding `vim.docs`. I can see it being useful. On our side, obviously, for adding documentation for plugins. Even as a user, were it exposed, it might be good as it could still generate a man page, even if it doesn't go to ndg. This would allow for self-documentation when users expose their nvf config as a package, for example (if we do decide to make it generate a man page). 

In the meantime, I don't see blockers for this PR if you want to merge and handle docs in another one, so I'll drop an approval

modules/neovim/init/filetype.nix
```diff
@@ -0,0 +1,121 @@
...
+            elseif ext == 'rst' then
+              return 'rst'
+            end
+          end,
+        '';
+      };
+
+      extension = {
+        mdx = "markdown";
+        bar = lib.mkLuaInline ''
```
horriblename: lib.generators.mkLuaInline

modules/neovim/init/filetype.nix
```diff
@@ -0,0 +1,121 @@
...
+          type = ints.unsigned;
+          description = ''
+            `vim.filetype.add()` can take an optional priority value to resolve
+            conflicts where a filetype is registered by multiple patterns. When
+            priority is specified, file with the higher priority value will be
+            matched first on conflict.
+          '';
+        };
+      };
+    })))
```
horriblename: - the first element should be a luaInline or str (it accepts function)

```
listOf (either (either str luaInline) (submodule {priority
```
 
- (optional) a custom tuple type would be good here, to prevent bogus values like `["dosini" "oops"]`. Not that important tho
  ```
  type = customTupleType [(either str luaInline) (submodule {priority...})]
  ```

modules/neovim/init/filetype.nix
```diff
@@ -0,0 +1,121 @@
...
+        bar = lib.mkLuaInline ''
+          bar = function(path, bufnr)
+            if some_condition() then
+              return 'barscript', function(bufnr)
+                -- Set a buffer variable
+                vim.b[bufnr].barscript_version = 2
+              end
+            end
+            return 'bar'
+          end,
```
horriblename:
```suggestion
        {
          bar = function(path, bufnr)
            if some_condition() then
              return 'barscript', function(bufnr)
                -- Set a buffer variable
                vim.b[bufnr].barscript_version = 2
              end
            end
            return 'bar'
          end,
         }
```

luaInline items should be valid lua expressions (missing braces)

horriblename (changes requested):
overall looks good, I don't really get what vim.doc is suppose to do so I'd like you to elaborate a bit

i.e. why not just add .md pages in the manual

NotAShelf:
> overall looks good, I don't really get what vim.doc is suppose to do so I'd like you to elaborate a bit
> 
> i.e. why not just add .md pages in the manual

In Nixpkgs, the modules can specify documentation bits under `meta.doc` to be included in the manual. I have been meaning to introduce something of the same kind in our module system, but in hindsight we could also just... parse `meta.doc`? That said `vim.docs` could still be a module that controls whether manpages are included in the closure.

NotAShelf:
Still not ready for a merge, but I'd appreciate an updated review.

modules/neovim/init/filetype.nix
```diff
@@ -0,0 +1,123 @@
...
+
+  # vim.filetype.add() is quite robust, but this makes for a very
+  # complex type that we have to handle. It takes a string, a Lua function
+  # or a dictionary with the priority of the extension.
+  ftOptionType = attrsOf (oneOf [
+    str # "filetype"
+    luaInline # `function(path, bufnr) ... end`
+
+    # { 'dosini', { priority = 10 } },
+    (listOf (either (either str luaInline) (submodule (submodule {
```
horriblename: is double submodule `submodule (submodule {` intended here?

horriblename: is that a yes?? 😭 

also, (either (either x y) z) should be `oneOf [x y z]`

horriblename (changes requested):

modules/neovim/init/mappings.nix
```diff
@@ -0,0 +1,156 @@
...
+  inherit (lib.options) mkOption literalMD;
+  inherit (lib.types) either str listOf attrsOf nullOr submodule;
+  inherit (lib.attrsets) mapAttrsToList;
+  inherit (lib.lists) flatten;
+  inherit (lib.trivial) pipe;
+  inherit (lib.options) mkEnableOption;
+  inherit (lib.nvim.config) mkBool;
+
+  mapConfigOptions = {
+    desc = mkOption {
```
snoweuph, line 16: can we also get a way to define the icon for which key?
Currently you can't really configure them via NVF, even though its supported by whichkey

snoweuph (approved):
LGTM

snoweuph (changes requested):
Still needs a Changelog...

modules/neovim/init/mappings.nix
NotAShelf: Not in the scope of this PR.

snoweuph (approved):
LGTM

horriblename (approved):

PR #1376, opened 2026-01-30
language/go: add `golangci-lint` for extra diagnostics

snoweuph:
Adds `golanci-lint` for more diagnostics in go.

works just fine with severity mapping of the config file:
image
image
image

## Sanity Checking

- Style and consistency
- Package(s) built:
  - [x] `.#maximal`
- Tested on platform(s)
  - [x] `x86_64-linux`
  - [ ] `aarch64-linux`
  - [ ] `x86_64-darwin`
  - [ ] `aarch64-darwin`

---

### DIFF
--- a/modules/plugins/languages/go.nix
+++ b/modules/plugins/languages/go.nix
@@ -10,7 +10,7 @@
   inherit (lib.meta) getExe;
   inherit (lib.generators) mkLuaInline;
   inherit (lib.types) bool enum package;
-  inherit (lib.nvim.types) mkGrammarOption deprecatedSingleOrListOf;
+  inherit (lib.nvim.types) mkGrammarOption diagnostics deprecatedSingleOrListOf;
   inherit (lib.nvim.dag) entryAfter;
   inherit (lib.nvim.attrsets) mapListToAttrs;
 
@@ -78,6 +78,91 @@
       package = pkgs.delve;
     };
   };
+
+  defaultDiagnosticsProvider = ["golangci-lint"];
+  diagnosticsProviders = {
+    golangci-lint = let
+      pkg = pkgs.golangci-lint;
+    in {
+      package = pkg;
+      config = {
+        cmd = getExe pkg;
+        args = [
+          "run"
+          "--output.json.path=stdout"
+          "--issues-exit-code=0"
+          "--show-stats=false"
+          "--fix=false"
+          "--path-mode=abs"
+          # Overwrite values that could be configured and result in unwanted writes
+          "--output.text.path="
+          "--output.tab.path="
+          "--output.html.path="
+          "--output.checkstyle.path="
+          "--output.code-climate.path="
+          "--output.junit-xml.path="
+          "--output.teamcity.path="
+          "--output.sarif.path="
+        ];
+        parser = mkLuaInline ''
+          function(output, bufnr)
+            local SOURCE = "golangci-lint";
+
+            local function display_tool_error(msg)
+              return{
+                {
+                  bufnr = bufnr,
+                  lnum = 0,
+                  col = 0,
+                  message = string.format("[%s] %s", SOURCE, msg),
+                  severity = vim.diagnostic.severity.ERROR,
+                  source = SOURCE,
+                },
+              }
+            end
+
+            if output == "" then
+              return display_tool_error("no output provided")
+            end
+
+            local ok, decoded = pcall(vim.json.decode, output)
+            if not ok then
+              return display_tool_error("failed to parse JSON output")
+            end
+
+            if not decoded or not decoded.Issues then
+              return display_tool_error("unexpected output format")
+            end
+
+            local severity_map = {
+              error   = vim.diagnostic.severity.ERROR,
+              warning = vim.diagnostic.severity.WARN,
+              info    = vim.diagnostic.severity.INFO,
+              hint    = vim.diagnostic.severity.HINT,
+            }
+            local diagnostics = {}
+            for _, issue in ipairs(decoded.Issues) do
+              local sev = vim.diagnostic.severity.ERROR
+              if issue.Severity and issue.Severity ~= "" then
+                local normalized = issue.Severity:lower()
+                sev = severity_map[normalized] or vim.diagnostic.severity.ERROR
+              end
+              table.insert(diagnostics, {
+                bufnr = bufnr,
+                lnum = issue.Pos.Line - 1,
+                col = issue.Pos.Column - 1,
+                message = issue.Text,
+                code = issue.FromLinter,
+                severity = sev,
+                source = SOURCE,
+              })
+            end
+            return diagnostics
+          end
+        '';
+      };
+    };
+  };
 in {
   options.vim.languages.go = {
     enable = mkEnableOption "Go language support";
@@ -134,6 +219,14 @@ in {
         default = debuggers.${cfg.dap.debugger}.package;
       };
     };
+    extraDiagnostics = {
+      enable = mkEnableOption "extra Go diagnostics" // {default = config.vim.languages.enableExtraDiagnostics;};
+      types = diagnostics {
+        langDesc = "Go";
+        inherit diagnosticsProviders;
+        inherit defaultDiagnosticsProvider;
+      };
+    };
   };
 
   config = mkIf cfg.enable (mkMerge [
@@ -177,6 +270,16 @@ in {
           }
         '';
         debugger.nvim-dap.enable = true;
+      };
+    })
+
+    (mkIf cfg.extraDiagnostics.enable {
+      vim.diagnostics.nvim-lint = {
+        enable = true;
+        linters_by_ft.go = cfg.extraDiagnostics.types;
+        linters =
+          mkMerge (map (name: {${name} = diagnosticsProviders.${name}.config;})
+            cfg.extraDiagnostics.types);
       };
     })
   ]);
--- a/docs/manual/release-notes/rl-0.9.md
+++ b/docs/manual/release-notes/rl-0.9.md
@@ -164,6 +164,8 @@
 - Added [tera](https://keats.github.io/tera/) language support (syntax
   highlighting only).
 
+- Added [`golangci-lint`](https://golangci-lint.run/) for more diagnostics.
+
 [vagahbond](https://github.com/vagahbond): [codewindow.nvim]:
 https://github.com/gorbit99/codewindow.nvim
 


modules/plugins/languages/go.nix
```diff
@@ -21,41 +21,46 @@
...
-
-          local function get_root(fname)
-            if _G.nvf_gopls_mod_cache and fname:sub(1, #_G.nvf_gopls_mod_cache) == _G.nvf_gopls_mod_cache then
-              local clients = vim.lsp.get_clients { name = 'gopls' }
-              if #clients > 0 then
-                return clients[#clients].config.root_dir
+      root_dir =
+        mkLuaInline
+        /*
+        lua
```
snoweuph: adding those comments results in syntax highlighting.
the weird breakover is just alejandra.

modules/plugins/languages/go.nix
```diff
@@ -78,6 +83,86 @@
...
+          "--output.json.path=stdout"
+          "--issues-exit-code=0"
+          "--show-stats=false"
+          "--fix=false"
+          "--path-mode=abs"
+        ];
+        parser =
+          mkLuaInline
+          /*
+          lua
```
snoweuph: adding those comments results in syntax highlighting.
the weird breakover is just alejandra.

modules/plugins/languages/go.nix
```diff
@@ -92,7 +177,7 @@ in {
       enable = mkEnableOption "Go LSP support" // {default = config.vim.lsp.enable;};
 
       servers = mkOption {
-        type = deprecatedSingleOrListOf "vim.language.go.lsp.servers" (enum (attrNames servers));
+        type = listOf (enum (attrNames servers));
```
snoweuph: thought i might just clean them up at the same time

modules/plugins/languages/go.nix
```diff
@@ -110,7 +195,7 @@ in {
 
       type = mkOption {
         description = "Go formatter to use";
-        type = deprecatedSingleOrListOf "vim.language.go.format.type" (enum (attrNames formats));
+        type = listOf (enum (attrNames formats));
```
snoweuph: thought i might just clean them up at the same time

snoweuph:
I plan on adding a few more common go diagnostic tools in followup MRs.
Is there a timeline, when the 0.9 release will be cut?
I would like to get them all into it :D

snoweuph:
> ### 🚀 Live preview deployed from [fde1338](https://github.com/NotAShelf/nvf/commit/fde1338793aeb88810122030cf0badd3297936b0)
> 
> View it [here](https:///nvf.notashelf.dev/docs-preview-1376/):
> **Debug Information**

image

modules/plugins/languages/go.nix
horriblename: don't do this. I know we have a few of those lying around but they're there either from before we decided against them, or it slipped past reviews.

(we don't like them simply because  alejandra formats them weirdly) 

there's a PR for custom injections that I'll try to get merged next week

modules/plugins/languages/go.nix
horriblename: we should keep these around till next release

modules/plugins/languages/go.nix
```diff
@@ -78,6 +83,86 @@
       package = pkgs.delve;
     };
   };
+
+  defaultDiagnosticsProvider = ["golangci-lint"];
+  diagnosticsProviders = {
+    golangci-lint = let
```
horriblename, line 84: is there any reason not to use `golangcilint` which nvim-lint ships by default?

horriblename:
> Is there a timeline, when the 0.9 release will be cut?


not anytime soon, we usually only make a release when we have breaking changes, and we don't have any planned at the moment (unless @NotAShelf has a release schedule he didn't tell me about)

modules/plugins/languages/go.nix
snoweuph: i didnt knew that nvim-lint already packages `golangcilint`. let me check it out real quick

snoweuph: Their package doesnt handle the Severites like configured

image


this should be an info not a warn.
Also their diags don't include the rule name, so its harder to configure just from neovim, as you don't know the correct codes.

modules/plugins/languages/go.nix
```diff
@@ -78,6 +78,81 @@
...
+      config = {
+        cmd = getExe pkg;
+        args = [
+          "run"
+          "--output.json.path=stdout"
+          "--issues-exit-code=0"
+          "--show-stats=false"
+          "--fix=false"
+          "--path-mode=abs"
+        ];
```
horriblename, line 106: nvim-lint's args covers some edge cases, either add those in or unset it to use nvim-lint values

```
  return {
    'run',
    '--output.json.path=stdout',
    -- Overwrite values possibly set in .golangci.yml
    '--output.text.path=',
    '--output.tab.path=',
    '--output.html.path=',
    '--output.checkstyle.path=',
    '--output.code-climate.path=',
    '--output.junit-xml.path=',
    '--output.teamcity.path=',
    '--output.sarif.path=',
    '--issues-exit-code=0',
    '--show-stats=false',
    -- Get absolute path of the linted file
    '--path-mode=abs',
    function()
      return vim.fn.fnamemodify(vim.api.nvim_buf_get_name(0), ":h")
    end,
  }
```

modules/plugins/languages/go.nix
```diff
@@ -78,6 +78,81 @@
       package = pkgs.delve;
     };
   };
+
+  defaultDiagnosticsProvider = ["golangci-lint"];
+  diagnosticsProviders = {
+    golangci-lint = let
```
horriblename, line 84: I think you should use the name `golangcilint`, to avoid confusion

modules/plugins/languages/go.nix
```diff
@@ -78,6 +78,81 @@
...
+                  col = 0,
+                  message = string.format("[%s] %s", SOURCE, msg),
+                  severity = vim.diagnostic.severity.ERROR,
+                  source = SOURCE,
+                },
+              }
+            end
+
+            if not output or output == "" then
+              return error("no output provided")
```
horriblename: I don't think you're suppose to throw errors in the parser function, just return {}

modules/plugins/languages/go.nix
```diff
@@ -78,6 +78,81 @@
...
+                  lnum = 0,
+                  col = 0,
+                  message = string.format("[%s] %s", SOURCE, msg),
+                  severity = vim.diagnostic.severity.ERROR,
+                  source = SOURCE,
+                },
+              }
+            end
+
+            if not output or output == "" then
```
horriblename: `not output` is not possible according to their type signature

modules/plugins/languages/go.nix
```diff
@@ -78,6 +78,81 @@
...
+              return error("no output provided")
+            end
+
+            local ok, decoded = pcall(vim.json.decode, output)
+            if not ok then
+              return error("failed to parse JSON output")
```
horriblename: again, don't throw, return {}

modules/plugins/languages/go.nix
```diff
@@ -78,6 +78,81 @@
...
+            end
+
+            if not decoded or not decoded.Issues then
+              return error("unexpected output format")
```
horriblename: return {}

modules/plugins/languages/go.nix
```diff
@@ -78,6 +78,81 @@
...
+          function(output, bufnr)
+            local SOURCE = "golangci-lint";
+
+            local function error(msg)
+              return{
+                {
+                  bufnr = bufnr,
+                  lnum = 0,
+                  col = 0,
+                  message = string.format("[%s] %s", SOURCE, msg),
```
horriblename, line 117: The source should not be printed in message, the formatting is handled by whoever is printing that message (usually the float option in vim.diagnostic.config)

modules/plugins/languages/go.nix
```diff
@@ -78,6 +78,81 @@
...
+                sev = severity_map[normalized] or vim.diagnostic.severity.ERROR
+              end
+              table.insert(diagnostics, {
+                bufnr = bufnr,
+                lnum = issue.Pos.Line - 1,
+                col = issue.Pos.Column - 1,
+                message = string.format("%s [%s]", issue.Text, issue.FromLinter),
+                severity = sev,
+                source = SOURCE,
+              })
```
horriblename: upstream uses 

```
message = issue.Text,
source = issue.FromLinter,
```

I'll leave it up to you if you want to keep your version or theirs

horriblename (changes requested):

modules/plugins/languages/go.nix
snoweuph: `golangci-lint` is the offical name. i think its much weirder that `nvim-lint` doesn't use the correct name. as somebody writing go code, i feel `golangcilint` is more confusing.

modules/plugins/languages/go.nix
snoweuph: the source is `golangci-lint`. `FromLinter` is actually the rule name in `golangci-lint` like `ineffasign`

modules/plugins/languages/go.nix
snoweuph: don't we wanna see why its not working, if its not working?

My Experience is: failing loud is better than silent.
**but if you don't want that in nvf, and prefer diagnostic tools failing silent, I'm fine with that. decision is on your side**

also I'm just displaying the errors on line 0 instead of flodding the vim error logging api.

out of my experince before using nvf, basicly all none-ls modules do that.

modules/plugins/languages/go.nix
snoweuph: <https://github.com/NotAShelf/nvf/pull/1376#discussion_r2749575576>

modules/plugins/languages/go.nix
snoweuph: <https://github.com/NotAShelf/nvf/pull/1376#discussion_r2749575576>

modules/plugins/languages/go.nix
snoweuph: This is the error handler from before. if theres an error, i as a dev would prefer knowing its context, thats why I included it only in the error helper.

besides that, lets use that thread for talking about error handeling: <https://github.com/NotAShelf/nvf/pull/1376#discussion_r2749575576>

modules/plugins/languages/go.nix
snoweuph: it isn't?

i just wrote lua, so expected that I should handle all cases., but if you say its handled upstream, i gonna remove that check

modules/plugins/languages/go.nix
horriblename: ok fair point but name it something else not `error`

snoweuph: gotcha

snoweuph: im naming it `display_tool_error`

horriblename: naming it `error` makes it look like you're throwing an error

snoweuph: I mean, I'm kindof.  I'm displaying the error of the tool as a diagnostics error on line 0.

or would you say `display_tool_error` is a @horriblename ?

horriblename: > or would you say display_tool_error is a @horriblename ? 

xd

nah either gh or my internet died for 5m so the msg got sent late

snoweuph: ahh GG

snoweuph: so I can mark this as resolved?

snoweuph: or i guess you just did (y)

modules/plugins/languages/go.nix
horriblename: it might be good to set the code field to FromLinter then. see `:h vim.Diagnostic`

again I'll leave the decision up to you, I might be missing some context

snoweuph:
i gonna test the changes i id again, just to be sure

modules/plugins/languages/go.nix
horriblename: if you do end up setting code = FromLinter don't print it in the message to prevent duplication as, by default the code is printed

snoweuph:
still works as expected.
@horriblename ready for re-review

modules/plugins/languages/go.nix
horriblename: did you see the above comments?

snoweuph: now I did.

applied this, works perfectly now.
I did oversee that the field code exists.

image

snoweuph:
this is a great learning experince for me :D

adding a few of the other missing diags from [my list](https://git.euph.dev/snoweuph/nvim#to-do), should go much faster, in the future :D

horriblename (approved):
thanks!

PR #1384, opened 2026-02-02
Keymap/extra options

irobot:
Currently, the options passed into `vim.keymaps` are sanitized to allow only a predefined set of options.
This commit adds the following option~~s~~ to that set:

- `remap` - by default mappings set via `vim.keymap.set` are `:noremap`
(see https://neovim.io/doc/user/lua-guide.html#_creating-mappings). This
option allows for the default noremap to be undone.
- ~~`buffer` - applies the mapping only the the specified buffer~~

Note that `remap` is marked as internal. It can instead be accessed as `vim.keymaps.*.noremap = false`.

## Sanity Checking

- [X] I have updated the [changelog] as per my changes
- [X] I have tested, and self-reviewed my code
- [X] My changes fit guidelines found in [hacking nvf]
- Style and consistency
  - [X] I ran **Alejandra** to format my code (`nix fmt`)
  - [X] My code conforms to the [editorconfig] configuration of the project
  - [X] My changes are consistent with the rest of the codebase
- Package(s) built:
  - [ ] `.#maximal`
- Tested on platform(s)
  - [X] `x86_64-linux`
  - [ ] `aarch64-linux`
  - [ ] `x86_64-darwin`
  - [ ] `aarch64-darwin`

---

### DIFF
--- a/configuration.nix
+++ b/configuration.nix
@@ -92,6 +92,7 @@ isMaximal: {
       ruby.enable = false;
       fsharp.enable = false;
       just.enable = false;
+      make.enable = false;
       qml.enable = false;
       jinja.enable = false;
       tailwind.enable = false;
--- a/docs/manual/release-notes/rl-0.9.md
+++ b/docs/manual/release-notes/rl-0.9.md
@@ -172,6 +172,8 @@
 
 - Added [`golangci-lint`](https://golangci-lint.run/) for more diagnostics.
 
+- Added Makefile support via `languages.make`.
+
 [vagahbond](https://github.com/vagahbond): [codewindow.nvim]:
 https://github.com/gorbit99/codewindow.nvim
 
--- a/modules/plugins/languages/default.nix
+++ b/modules/plugins/languages/default.nix
@@ -52,6 +52,7 @@ in {
     ./yaml.nix
     ./ruby.nix
     ./just.nix
+    ./make.nix
     ./xml.nix
 
     # This is now a hard deprecation.
--- a/modules/plugins/languages/make.nix
+++ b/modules/plugins/languages/make.nix
@@ -1,0 +1,93 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}: let
+  inherit (builtins) attrNames;
+  inherit (lib.options) mkEnableOption mkOption;
+  inherit (lib.meta) getExe;
+  inherit (lib.types) listOf enum;
+  inherit (lib.modules) mkIf mkMerge;
+  inherit (lib.nvim.types) mkGrammarOption diagnostics;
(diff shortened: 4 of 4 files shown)


horriblename:
thank you for your efforts, however there are several issues with the new options 

- buffer is usually used in a FileType event or similar to add keymaps that only exist in specific types of buffers, it doesn't make sense to include it in our API, which runs once at startup hence there is no way to bind to a buffer in a useful way

- remap isn't implemented correctly. I'm not sure what you can do to fix it, but making `noremap` do nothing is 1. a breaking change, and 2. a terribly unintuitive API

there are probably ways you can add the remap option in a way that plays nice with noremap but I don't think it's worth your time

horriblename:
it seems noremap is broken/not documented correctly: `https://github.com/neovim/neovim/issues/37676`

ok, so the easiest fix right now for your problem, is to keep your new remap option, set it to `not noremap`, and mark it as private (`mkOption {internal = true;}`)

the reasoning for marking it private is because, I don't want to keep both `noremap` and `remap` as it is imo more annoying to keep both options than it's worth; but I'd like to wait for a response from neovim maintainers regarding the above issue before deciding whether to keep noremap or remap

modules/neovim/mappings/options.nix
```diff
@@ -25,6 +25,12 @@
     expr = mkBool false "Means that the action is actually an expression. Equivalent to adding <expr> to a map.";
     unique = mkBool false "Whether to fail if the map is already defined. Equivalent to adding <unique> to a map.";
     noremap = mkBool true "Whether to use the 'noremap' variant of the command, ignoring any custom mappings on the defined action. It is highly advised to keep this on, which is the default.";
+    remap = mkOption {
+      default = false;
+      type = bool;
+      internal = true;
+      description = "'noremap' is on by default. This option is used internally when noremap is set to false.";
+    };
```
horriblename: just realized my suggestion isn't rly necessary , you can remove this option; setting the config in the other file is enough

modules/wrapper/rc/config.nix
```diff
@@ -37,9 +37,11 @@ in {
...
       else keymap.action;
 
-    getOpts = keymap: {
-      inherit (keymap) desc silent nowait script expr unique noremap;
-    };
+    getOpts = keymap:
+      {
+        inherit (keymap) desc silent nowait script expr unique noremap;
+      }
+      // {remap = !keymap.noremap;};
```
horriblename:
```suggestion
        inherit (keymap) desc silent nowait script expr unique noremap;
        remap = !keymap.noremap;
     }
```

unnecessary `//`

horriblename (changes requested):

PR #1446, opened 2026-03-12
various: make `mkPackageOption` nullable for Treesitter grammars;  filter null grammars

NotAShelf:
We now allow `null`ing grammar packages that use `mkGrammarOption` to allow toggling of misc grammars that are not critical to a language module. `null` values are filtered out before building plugins. This hopefully provides greater flexibility in language module configuration, and prevents errors from `null` grammars.


horriblename (approved):
LGTM thanks

Issue #718, opened 2025-03-17
Duplicate Keybindings when setting personal keymaps in which-key


overtoneblue:
Any way to disable the default mapping that nvf is sending to which-key? it seems that if I want to set my own binds this then results in duplicate keybindings when they are not needed, one that is the default and one that I set personally.

NotAShelf:
@horriblename and I have discussed this recently. In a future release, chances are we'll add a top-level toggle (or multiple) to
1. enable vendored keybinds
2. enable which-key default

Unfortunately this is not yet real. Until the, I suggest using `vim.binds.whichKey.register` manually and forcing your own value.

PR #1450, opened 2026-03-15
Allow disabling nvf mappings

alfarelcynthesis:
This approach to the problem sources `mkMappingOption` from `config`, since it should reference an option for whether to use the provided default or `null`.

This change would be completely transparent to any current users, but requires enforcing that new plugin contributions use `config.vim.lib.mkMappingOption` instead of the `lib.nvim.binds` version.
If people are using the original `mkMappingOption` in outside projects, it will continue working as before, which is probably the desired behaviour.

- First commit contains the actual implementation.
  - Probably best to focus review time here initially, it's fairly simple.
- Second is mostly script-generated, replaces references to the original `mkMappingOption` with ones to the config-aware one.
- Third is fixes to less-automatable issues related to nullable keymaps.
- Fourth is conversions to `mkMappingOption` for options that should have used it but used to use `mkOption`.

Related discussion: #1043

## Feedback Wanted

- Should the lib options go elsewhere? be named something else?
- Should the config-aware `mkMappingOption` be named separately to reduce confusion?
- Should `useNvfKeymaps` be named something else?
- Should this be mentioned in the setup/other docs?
- [x] Should the option also be used to remove which-key group defaults (#718)?
   - Probably not, based on the comment on the issue.

## Remaining work
 
- [x] Direct testing. Packages build, but I should spend some more time inspecting them in an actual setup.
- [x] Make `surround-nvim` vendored keymaps option default depend on the global setting?
- [x] LS keybinds, does rust still require them to be not null?
- [x] Should the `copilot-lua` `mkLuaKeymap` use `lib.binds.mkLuaBinding`?
   - Updated it to allow nulls, should be fine to leave as is, but might be a good opportunity to update it?
- [x] Check that `.#nix` and `.#maximal` build with `useNvfKeymaps` disabled.

## Sanity Checking

  - I haven't tested every plugin, but the changes **should** be completely transparent, and I've tested both packages with and without the option enabled, as well as my config.
    This is probably worth testing a bit more thoroughly, if it gets accepted.
- Style and consistency
  - [ ] My changes are consistent with the rest of the codebase
    - debatable, worth talking about (function as config option)
    - isn't breaking
- Package(s) built:
  - [x] `.#maximal`
     - unrelated errors
- Tested on platform(s)
  - [x] `x86_64-linux`
  - [ ] `aarch64-linux`
  - [ ] `x86_64-darwin`
  - [ ] `aarch64-darwin`

---

### DIFF
--- a/modules/neovim/mappings/options.nix
+++ b/modules/neovim/mappings/options.nix
@@ -1,5 +1,5 @@
 {lib, ...}: let
-  inherit (lib.options) mkOption literalMD;
+  inherit (lib.options) mkEnableOption mkOption literalMD;
   inherit (lib.types) either str listOf attrsOf nullOr submodule;
   inherit (lib.nvim.config) mkBool;
 
@@ -57,6 +57,13 @@
     };
 in {
   options.vim = {
+    vendoredKeymaps.enable =
+      mkEnableOption "this project's vendored keymaps by default"
+      // {
+        default = true;
+        example = false;
+      };
+
     keymaps = mkOption {
       type = listOf mapType;
       description = "Custom keybindings.";
--- a/docs/manual/configuring/keybinds.md
+++ b/docs/manual/configuring/keybinds.md
@@ -1,7 +1,9 @@
 # Custom keymaps {#ch-keymaps}
 
-Some plugin modules provide keymap options for your convenience. If a keymap is
-not provided by such module options, you may easily register your own custom
+Some plugin modules provide keymap options for your convenience. These can be
+disabled by toggling {option}`vim.vendoredKeymaps.enable`. It is also possible
+to disable individual keymaps with options by setting them to `null`. If a
+keymap is not provided by a module, you may easily register your own custom
 keymaps via {option}`vim.keymaps`.
 
 ```nix
@@ -22,7 +24,7 @@ keymaps via {option}`vim.keymaps`.
     {
       key = "<leader>k";
       mode = ["n" "x"];
-      
+
       # While `lua` is `true`, `action` is expected to be
       # a valid Lua expression.
       lua = true;
--- a/docs/manual/hacking.md
+++ b/docs/manual/hacking.md
@@ -768,7 +768,7 @@ usage should look something like this:
 # pluginDefinition.nix
 {lib, ...}: let
   inherit (lib.options) mkEnableOption;
-  inherit (lib.nvim.binds) mkMappingOption;
+  inherit (config.vim.lib) mkMappingOption;
 in {
   options.vim.plugin = {
     enable = mkEnableOption "Enable plugin";
--- a/docs/manual/release-notes/rl-0.9.md
+++ b/docs/manual/release-notes/rl-0.9.md
@@ -195,6 +195,7 @@
     {command}`:healthcheck` doesn't know that.
   - Remove [which-key.nvim] `<leader>o` `+Notes` description which did not
     actually correspond to any keybinds.
+- Allow disabling nvf's vendored keymaps by toggling `vendoredKeymaps.enable`.
 
 [pyrox0](https://github.com/pyrox0):
 
--- a/modules/modules.nix
+++ b/modules/modules.nix
@@ -57,6 +57,7 @@
       "rc"
       "warnings"
       "lazy"
+      "lib"
     ];
 
     # Extra modules, such as deprecation warnings
--- a/modules/plugins/assistant/chatgpt/chatgpt.nix
+++ b/modules/plugins/assistant/chatgpt/chatgpt.nix
@@ -1,7 +1,11 @@
-{lib, ...}: let
+{
+  config,
+  lib,
+  ...
+}: let
   inherit (lib.options) mkEnableOption;
-  inherit (lib.nvim.binds) mkMappingOption;
   inherit (lib.nvim.types) mkPluginSetupOption;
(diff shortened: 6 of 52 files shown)


horriblename (dismissed):
Thanks! I wouldn't have thought of that

LGTM but I'd like an approval from @NotAShelf or @Soliprem

snoweuph:
> Thanks! I wouldn't have thought of that
> 
> LGTM but I'd like an approval from @NotAShelf or @Soliprem

isn't raf on for vacation according to matrix?

NotAShelf:
I'm on """vacation""", meaning I am away from my main system involuntarily

snoweuph:
> I'm on """vacation""", meaning I am away from my main system involuntarily

LMFAO, :3

alfarelcynthesis:
I added the changelog entry, seems like this won't need too many other changes :)

alfarelcynthesis:
Rebased on latest main ~~+ #1457~~.

(I'm rebasing to make sure to catch small incompatibilities as they get into the repo, given this touches a lot of files.)

(Also means anyone testing this branch by using it gets up-to-date options elsewhere.)

Soliprem (commented):
on a skim nothing jumps to the eye that I'd change, and I like the change itslef. LGTM, but I'll need to review this better with a fresh mind

NotAShelf (commented):
This is an interesting solution. I'm quite positive that this is the only way of scoping config and lib together, but I'm also not a huge fan of putting `lib` inside config. That said, I also cannot think of a better alternative so I'm not going to block this any further. 

One thing I want to note is that `useNvfKeymaps` could probably be renamed to something like `vim.keymaps.builtin.enable`. I like this more with the possibility of renaming this project (if ever) and the possibility of adding more knobs for builtin keymaps in mind. Also open to different names idea if you think `vim.keymaps.builtin` should not be reserved like that.

Left one comment, but it's not a blocker. You can get that while renaming the option if you'd like, but not critical. I'd also like to mention this in the rendered manual, a quick entry (as a note/important level admonition) would be perfectly okay in the "Custom keymaps" section I think.

Thank you for your good work, and your patience so far.

horriblename:
> One thing I want to note is that useNvfKeymaps could probably be renamed to something like `vim.keymaps.builtin.enable`

builtin is a bit ambiguous imo - builtin to neovim or builtin to nvf? `vendored` is a bit better, but also open to better names

alfarelcynthesis:
Unfortunately, since `vim.keymaps` is a list type, I can't add options there directly. Maybe `vim.vendoredKeymaps` is better? I don't think adding to the legacy `maps` is a good idea.

Honestly, all the nvf-specific one-off options might be better placed in a specific attrset so there aren't conflicts like this, but that would be annoying migration given most of them work fine as they are.

For now, I just updated the option to remove `nvf` from the name, but I'm of course up to change it if there's ideas.

alfarelcynthesis:
> This is an interesting solution. I'm quite positive that this is the only way of scoping config and lib together, but I'm also not a huge fan of putting `lib` inside config. That said, I also cannot think of a better alternative so I'm not going to block this any further.
> 
> One thing I want to note is that `useNvfKeymaps` could probably be renamed to something like `vim.keymaps.builtin.enable`. I like this more with the possibility of renaming this project (if ever) and the possibility of adding more knobs for builtin keymaps in mind. Also open to different names idea if you think `vim.keymaps.builtin` should not be reserved like that.
> 
> Left one comment, but it's not a blocker. You can get that while renaming the option if you'd like, but not critical. I'd also like to mention this in the rendered manual, a quick entry (as a note/important level admonition) would be perfectly okay in the "Custom keymaps" section I think.
> 
> Thank you for your good work, and your patience so far.

You mention a (presumably separate) comment here but I can't seem to find it. Do you remember what it was that you preferred changed (other than the option name and a doc hint)?

alfarelcynthesis:
Also, I've added a couple lines to the `Custom keymaps` section, ~~but looking at the rendered docs I can't actually find the section.~~
(Found it...)

modules/neovim/mappings/options.nix
```diff
@@ -57,6 +57,8 @@
     };
 in {
   options.vim = {
+    vendoredKeymaps = mkEnableOption "this project's vendored keymaps by default" // {default = true;};
```
NotAShelf: This should be `vim.vendoredKeyboards.enable` for the sake of allowing future extensibility. I cannot say for certain that this will happen, but it's critical that we prefer this kind of future-proof design imho.

NotAShelf (changes requested):
Thanks a lot for your patience.

One last change I'm requesting to avoid more headache down the line. Otherwise, I think this is ready to be merged with @Soliprem, @snoweuph and @horriblename's approvals.

modules/neovim/mappings/options.nix
alfarelcynthesis: Good idea, thanks for pointing that out. Should be fixed now.

alfarelcynthesis:
Noticed that I hadn't set the `mkEnableOption` example to false, so I added that as well.

snoweuph (approved):
LGTM, though I'm not a fan of the term "vendored" for the builtin/default provided keybindings.

though I understand that builtin and default aren't great terms either, as it could be confused with neovim builtins and defaults.

no better term comes to my mind, so I guess vendored is fine.

horriblename (approved):

NotAShelf (approved):

PR #1454, opened 2026-03-16
ci: use nvf cachix everywhere

horriblename:


.github/workflows/check.yml
```diff
@@ -84,6 +84,9 @@ jobs:
     runs-on: ubuntu-latest
     if: "!contains(github.event.pull_request.title, '[skip ci]')"
     strategy:
+      # avoid having 2 jobs building ndg when no cache is available
+      # TODO: there is probably a better way
+      max-parallel: 1
```
horriblename, line 89: @NotAShelf if it's okay with you, I'd like to get rid of the matrix strategy here and just do `nix build .#docs .#docs-html .#docs-manpages .#docs-json` so I can just not worry about redundant ndg builds in parallel

PR #569, opened 2025-01-19
Added Tex Language Support

isaacST08:
Added support for the tex language. 
Added the texlab LSP along with config options for build/compiling:
- Compiling package option
- Compiling executable option
- Compiling executable args option
- PDF preview using your PDF viewer of choice
- PDF preview update on save option

Not implemented (yet):
- Support for other LSPs
- Formatting support via nvf options
- Texlab options (There is however an `extraLuaSettings` option for texlab that allows setting these options in the meantime): 
  - chktex  
  -  diagnosticsDelay
  -  diagnostics
  - symbols
  - formatterLineLength
  - bibtexFormatter
  - latexFormatter
  - latexindent
  - completion
  - inlayHints
  - experimental

Defaults mostly follow the [texlab defaults](https://github.com/latex-lsp/texlab/wiki/Configuration) with some exceptions such as the default compiler being tectonic and the default file viewer being okular.

(I also alphabetically sorted the language imports in the modules/plugins/languages/default.nix)

PR #1455, opened 2026-03-16
Feature language tex

horriblename:
supercedes #569 

stripped out more complex options like builder and pdfViewer

reverted changes to use the fork of new-file-template.nvim - I think that should have been done as a plugin override by the user


horriblename:
closed in favor of #1465

PR #1514, opened 2026-04-11
LSP Refactoring

snoweuph:
Related: <https://github.com/NotAShelf/nvf/issues/1484>

Modules this won't update as they rely heavily on plugins and/or lua and would need to be completely rebuild.

- Scala
- C#
- Rust
- Haskell

Self tested on Following Languages:
- Nix
- Markdown
- Assembly
- Just 
- HTML
- JSON
- JavaScript
- TypeScript
- C
- Vala
- Go
- XML
- SCSS 

## Sanity Checking

[editorconfig]: https://editorconfig.org
[changelog]: https://github.com/NotAShelf/nvf/tree/main/docs/manual/release-notes
[hacking nvf]: https://nvf.notashelf.dev/hacking.html#sec-guidelines

- [x] I have updated the [changelog] as per my changes
- [x] I have tested, and self-reviewed my code
- [x] My changes fit guidelines found in [hacking nvf]
- Style and consistency
  - [x] I ran **Alejandra** to format my code (`nix fmt`)
  - [x] My code conforms to the [editorconfig] configuration of the project
  - [x] My changes are consistent with the rest of the codebase
- If new changes are particularly complex:
  - [x] My code includes comments in particularly complex areas
  - [ ] I have added a section in the manual
  - [x] _(For breaking changes)_ I have included a migration guide
- Package(s) built:
  - [x] `.#nix` _(default package)_
  - [x] `.#maximal`
  - [x] `.#docs-html` _(manual, must build)_
  - [x] `.#docs-linkcheck` _(optional, please build if adding links)_
- Tested on platform(s)
  - [x] `x86_64-linux`
  - [ ] `aarch64-linux`
  - [ ] `x86_64-darwin`
  - [ ] `aarch64-darwin`

---

Add a :+1: [reaction] to [pull requests you find important].

[reaction]: https://github.blog/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/
[pull requests you find important]: https://github.com/NixOS/nixpkgs/pulls?q=is%3Aopen+sort%3Areactions-%2B1-desc

Issue #1506, opened 2026-04-09
roslyn_ls is using a deprecated api


anispwyn:
### I have confirmed that this is a bug related to nvf

- [x] This is a bug, and not an user error or a support request. I understand that my issue will be closed if it is not a bug in nvf.
- [x] I have checked the [issues tab](https://github.com/notashelf/nvf/issues?q=is%3Aissue) and confirmed that my issue has not yet been reported. I understand that my issue will be closed if it is a duplicate.

### Description

roslyn keep throwing a api deprecation warnings, which when you execute "nvim" without any args, id throw the warning on your face and require you to press enter once

### Installation Method

Home Manager Module (`homeManagerModules.default`)

### nvf Version

master

### Reproduction steps

1. enable csharp languages (programs.nvf.settings.languages.csharp.enable)
2. enable rosyln_ls language server
3. deprecation warning

### Expected behavior

no deprecation warnings

### Actual Behavior

deprecation warning shows, which require pressing enter to ignore

### System Information

```bash
- system: `"x86_64-linux"`
 - host os: `Linux 6.19.11-cachyos-lto, NixOS, 26.05 (Yarara), 26.05.20260405.68d8aa3`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Lix, like Nix) 2.95.1
System type: x86_64-linux
Additional system types: i686-linux, x86_64-v1-linux, x86_64-v2-linux, x86_64-v3-linux
Features: gc, signed-caches
System configuration file: /etc/nix/nix.conf
User configuration files: /home/anisp/.config/nix/nix.conf:/nix/store/k4lq0r3d14xpf4r4y8q4kf63s49iii4p-stylix-kde-config/nix/nix.conf:/etc/xdg/nix/nix.conf:/home/anisp/.nix-profile/etc/xdg/nix/nix.conf:/nix/profile/etc/xdg/nix/nix.conf:/home/anisp/.local/state/nix/profile/etc/xdg/nix/nix.conf:/etc/profiles/per-user/anisp/etc/xdg/nix/nix.conf:/nix/var/nix/profiles/default/etc/xdg/nix/nix.conf:/run/current-system/sw/etc/xdg/nix/nix.conf
Store directory: /nix/store
State directory: /nix/var/nix
Data directory: /nix/store/m1srmdgh3bf4rd0wv0wcib7bqr9xxrwz-lix-2.95.1/share`
[...]

PR #1515, opened 2026-04-11
languages/csharp: update

CaueAnjos:
This updates the C# language server configurations to align with `nvim-lspconfig` and refactors `roslyn` into `roslyn-ls`. I believe this approach is more intuitive, avoiding the confusion of having two similarly named servers with nearly identical configurations.

## Sanity Checking

- Style and consistency
- Package(s) built:
  - [ ] `.#maximal`
- Tested on platform(s)
  - [x] `x86_64-linux`
  - [ ] `aarch64-linux`
  - [ ] `x86_64-darwin`
  - [ ] `aarch64-darwin`

---

### DIFF
--- a/modules/plugins/languages/csharp.nix
+++ b/modules/plugins/languages/csharp.nix
@@ -5,7 +5,6 @@
   options,
   ...
 }: let
-  inherit (builtins) concatMap;
   inherit (builtins) elem;
   inherit (lib) genAttrs;
   inherit (lib.generators) mkLuaInline;
@@ -18,13 +17,8 @@
   inherit (lib.nvim.dag) entryAnywhere;
   inherit (lib.nvim.binds) addDescriptionsToMappings;
 
-  extraServerPlugins = {
-    csharp_ls = ["csharpls-extended-lsp-nvim"];
-    roslyn-ls = [];
-    omnisharp = [];
-  };
-  defaultServers = ["csharp-ls"];
-  servers = ["csharp-ls" "omnisharp" "roslyn-ls"];
+  defaultServers = ["csharp_ls"];
+  servers = ["csharp_ls" "omnisharp" "roslyn-ls"];
 
   cfg = config.vim.languages.csharp;
 in {
@@ -35,7 +29,7 @@ in {
 
         ::: {.note}
         This feature will not work if the .NET SDK is not installed.
-        Both `roslyn-ls` (with `roslyn-nvim`) and `csharp-ls` require the .NET SDK to function properly with Razor.
+        Both `roslyn-ls` (with `roslyn-nvim`) and `csharp_ls` require the .NET SDK to function properly with Razor.
         Ensure that the .NET SDK is installed.
 
         Check for version compatibility for optimal performance.
@@ -43,7 +37,7 @@ in {
 
         ::: {.warning}
         At the moment, only `roslyn-ls`(with roslyn-nvim) provides full Razor support.
-        `csharp-ls` is limited to `.cshtml` files.
+        `csharp_ls` is limited to `.cshtml` files.
         :::
       '';
 
@@ -117,6 +111,15 @@ in {
             listReferences = mkMappingOption "List references [omnisharp-extended-lsp-nvim]" mappings.listReferences;
             listImplementations = mkMappingOption "List implementations [omnisharp-extended-lsp-nvim]" mappings.listImplementations;
           };
+        };
+        csharpls-extended-lsp-nvim = {
+          enable = mkEnableOption ''
+            Extended 'textDocument/definition' handler for csharp_ls Neovim LSP
+
+            ::: {.note}
+            This feature only works for `csharp_ls`.
+            :::
+          '';
         };
       };
 
@@ -208,6 +211,18 @@ in {
               ${mkBinding mappings.listImplementations "require('omnisharp_extended').lsp_implementation"}
             end
           '';
+      };
+    })
+    (mkIf (cfg.lsp.enable
+      && cfg.extensions.csharpls-extended-lsp-nvim.enable
+      && (elem "csharp_ls" cfg.lsp.servers)) {
+      vim = {
+        startPlugins = ["csharpls-extended-lsp-nvim"];
+        lsp.servers.csharp_ls.on_attach = mkLuaInline ''
+          function(client, bufnr)
+            require('csharpls_extended').buf_read_cmd_bind()
+          end
+        '';
       };
     })
   ]);
--- a/modules/plugins/languages/csharp.nix
+++ b/modules/plugins/languages/csharp.nix
@@ -2,201 +2,14 @@
   lib,
   pkgs,
   config,
-  options,
   ...
 }: let
-  inherit (builtins) attrNames concatMap;
+  inherit (builtins) concatMap;
+  inherit (lib) genAttrs;
   inherit (lib.options) mkEnableOption mkOption literalExpression;
-  inherit (lib.types) enum;
+  inherit (lib.types) enum listOf;
   inherit (lib.modules) mkIf mkMerge;
-  inherit (lib.meta) getExe;
-  inherit (lib.generators) mkLuaInline;
-  inherit (lib.strings) optionalString;
-  inherit (lib.nvim.types) mkGrammarOption mkPluginSetupOption deprecatedSingleOrListOf;
-  inherit (lib.nvim.lua) toLuaObject;
-  inherit (lib.nvim.attrsets) mapListToAttrs;
-
-  lspKeyConfig = config.vim.lsp.mappings;
-  lspKeyOptions = options.vim.lsp.mappings;
-  mkLspBinding = optionName: action: let
-    key = lspKeyConfig.${optionName};
-    desc = lspKeyOptions.${optionName}.description;
-  in
-    optionalString (key != null) "vim.keymap.set('n', '${key}', ${action}, {buffer=bufnr, noremap=true, silent=true, desc='${desc}'})";
-
-  defaultServers = ["csharp_ls"];
-  servers = {
-    omnisharp = {
-      cmd = mkLuaInline ''
-        {
-          ${toLuaObject (getExe pkgs.omnisharp-roslyn)},
-          '-z', -- https://github.com/OmniSharp/omnisharp-vscode/pull/4300
-          '--hostPID',
-          tostring(vim.fn.getpid()),
-          'DotNet:enablePackageRestore=false',
-          '--encoding',
-          'utf-8',
-          '--languageserver',
-        }
-      '';
-      filetypes = ["cs" "vb"];
-      root_dir = mkLuaInline ''
-        function(bufnr, on_dir)
-          local function find_root_pattern(fname, lua_pattern)
-            return vim.fs.root(0, function(name, path)
-              return name:match(lua_pattern)
-            end)
-          end
-
-          local fname = vim.api.nvim_buf_get_name(bufnr)
-          on_dir(find_root_pattern(fname, "%.sln$") or find_root_pattern(fname, "%.csproj$"))
-        end
-      '';
-      init_options = {};
-      capabilities = {
-        workspace = {
-          workspaceFolders = false; # https://github.com/OmniSharp/omnisharp-roslyn/issues/909
-        };
-      };
-      on_attach = mkLuaInline ''
-        function(client, bufnr)
-          local oe = require("omnisharp_extended")
-          ${mkLspBinding "goToDefinition" "oe.lsp_definition"}
-          ${mkLspBinding "goToType" "oe.lsp_type_definition"}
-          ${mkLspBinding "listReferences" "oe.lsp_references"}
-          ${mkLspBinding "listImplementations" "oe.lsp_implementation"}
-        end
-      '';
-      settings = {
-        FormattingOptions = {
-          # Enables support for reading code style, naming convention and analyzer
-          # settings from .editorconfig.
-          EnableEditorConfigSupport = true;
-          # Specifies whether 'using' directives should be grouped and sorted during
-          # document formatting.
-          OrganizeImports = null;
-        };
-        MsBuild = {
-          # If true, MSBuild project system will only load projects for files that
-          # were opened in the editor. This setting is useful for big C# codebases
-          # and allows for faster initialization of code navigation features only
-          # for projects that are relevant to code that is being edited. With this
-          # setting enabled OmniSharp may load fewer projects and may thus display
-          # incomplete reference lists for symbols.
-          LoadProjectsOnDemand = null;
-        };
-        RoslynExtensionsOptions = {
-          # Enables support for roslyn analyzers, code fixes and rulesets.
-          EnableAnalyzersSupport = null;
-          # Enables support for showing unimported types and unimported extension
-          # methods in completion lists. When committed, the appropriate using
-          # directive will be added at the top of the current file. This option can
-          # have a negative impact on initial completion responsiveness;
-          # particularly for the first few completion sessions after opening a
-          # solution.
-          EnableImportCompletion = null;
-          # Only run analyzers against open files when 'enableRoslynAnalyzers' is
-          # true
-          AnalyzeOpenDocumentsOnly = null;
-          # Enables the possibility to see the code in external nuget dependencies
-          EnableDecompilationSupport = null;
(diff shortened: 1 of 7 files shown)


snoweuph:
In #1514 we are isolating the LSPs into presets.

Due to limited understand on our side, how the LSPs for C# work, why theres so many plugins, just to get LSPs up and running, and which LSP is which (3x roslyn**) \
We are not touching the c# Module in #1514.

Would you be open to waiting with this after the #1514, and helping us isolate the LSPs from this clearly into presets?

CaueAnjos:
> In #1514 we are isolating the LSPs into presets.
> 
> Due to limited understand on our side, how the LSPs for C# work, why theres so many plugins, just to get LSPs up and running, and which LSP is which (3x roslyn**) We are not touching the c# Module in #1514.
> 
> Would you be open to waiting with this after the #1514, and helping us isolate the LSPs from this clearly into presets?

For sure

snoweuph:
#1514 is merged into main :3

snoweuph:
I guess this will fix <https://github.com/NotAShelf/nvf/issues/1506> ?

snoweuph:
@CaueAnjos you had alot of conflicts and did some weird stuff.

I rebased your commits to be on main without conflicts, hope that helps :3

modules/plugins/lsp/presets/omnisharp.nix
```diff
@@ -0,0 +1,97 @@
...
+  lib,
+  pkgs,
+  ...
+}: let
+  inherit (lib.meta) getExe;
+  inherit (lib.modules) mkIf;
+  inherit (lib.nvim.types) mkLspPresetEnableOption;
+  inherit (lib.generators) mkLuaInline;
+
+  cfg = config.vim.lsp.presets.omnisharp;
```
snoweuph, line 12: This should be called `omnisharp-roslyn` not `omnisharp` to match the name in nixpkgs.

modules/plugins/lsp/presets/roslyn_ls.nix
```diff
@@ -0,0 +1,402 @@
...
+  lib,
+  pkgs,
+  ...
+}: let
+  inherit (lib.meta) getExe;
+  inherit (lib.modules) mkIf;
+  inherit (lib.nvim.types) mkLspPresetEnableOption;
+  inherit (lib.generators) mkLuaInline;
+
+  cfg = config.vim.lsp.presets.roslyn_ls;
```
snoweuph: please  name it `roslyn-ls` like the <nixpkg>

modules/plugins/lsp/presets/csharp_ls.nix
```diff
@@ -0,0 +1,50 @@
...
+  lib,
+  pkgs,
+  ...
+}: let
+  inherit (lib.meta) getExe;
+  inherit (lib.modules) mkIf;
+  inherit (lib.nvim.types) mkLspPresetEnableOption;
+  inherit (lib.generators) mkLuaInline;
+
+  cfg = config.vim.lsp.presets.csharp_ls;
```
snoweuph, line 12: please call it `csharp-ls` like the <nixpkg>

modules/plugins/lsp/presets/omnisharp.nix
```diff
@@ -0,0 +1,97 @@
...
+in {
+  options.vim.lsp.presets.omnisharp = {
+    enable = mkLspPresetEnableOption "omnisharp" "C#" [];
```
snoweuph:
```suggestion
    enable = mkLspPresetEnableOption "omnisharp-roslyn" "Omnisharp Roslyn" [];
```

modules/plugins/lsp/presets/roslyn_ls.nix
```diff
@@ -0,0 +1,402 @@
...
+in {
+  options.vim.lsp.presets.roslyn_ls = {
+    enable = mkLspPresetEnableOption "roslyn_ls" "C#" [];
```
snoweuph:
```suggestion
    enable = mkLspPresetEnableOption "roslyn-ls" "Roslyn" [];
```

modules/plugins/lsp/presets/csharp_ls.nix
```diff
@@ -0,0 +1,51 @@
...
+in {
+  # HACK: this server shoud be named `csharp-ls`, but the extension `csharpls-extended-lsp-nvim` only works if it is named `csharp_ls`
+  options.vim.lsp.presets.csharp_ls = {
+    enable = mkLspPresetEnableOption "csharp_ls" "CSharp" [];
```
snoweuph: The 2nd one is the Displayname
```suggestion
    enable = mkLspPresetEnableOption "csharp_ls" "C#" [];
```

modules/plugins/languages/csharp.nix
```diff
@@ -5,206 +5,31 @@
...
-        {
-          ${toLuaObject (getExe pkgs.roslyn-ls)},
-          '--logLevel=Warning',
-          '--extensionLogDirectory=' .. vim.fs.dirname(vim.lsp.get_log_path()),
-          '--stdio',
-        }
-      '';
+  defaultFormat = ["lsp"];
+  formats = {
+    lsp = {};
```
snoweuph: by default conform.nvim will try using the LSP. please don't add this placeholder option. This doesn't match all the other lang modules.

Even when `format.enable = false` is set, and a user has an LSP registerd that supports formatting, it will format with the LSP.

That is default behaviour, so this option doesn't actually add any value as far as I see.

modules/plugins/languages/csharp.nix
```diff
@@ -232,11 +57,37 @@ in {
             Roslyn LSP plugin for neovim
```
snoweuph: Maybe add a comment explaining, what value this extension adds, like you did for `omnisharp-extended-lsp-nvim`. If there is none, we might be better of dropping the plugin.

modules/plugins/languages/csharp.nix
```diff
@@ -259,10 +110,25 @@ in {
           };
         servers = mkOption {
           description = "C# LSP server to use";
-          type = deprecatedSingleOrListOf "vim.language.csharp.lsp.servers" (enum (attrNames servers));
+          type = listOf (enum servers);
```
snoweuph:
```suggestion
          type = listOf (enumWithRename
            "vim.languages.csharp.lsp.servers"
            servers
            {
              roslyn_ls = "roslyn-ls";
            });
```

you will need to also add the import

```nix
let
 inherit (lib.nvim.types) enumWithRename;
in
```

modules/plugins/languages/csharp.nix
```diff
@@ -288,20 +153,112 @@ in {
...
+            then "vim.keymap.set('n', ${toLuaObject binding.value}, ${action}, {buffer=bufnr, noremap=true, silent=true, desc=${toLuaObject binding.description}})"
+            else "";
+        in [
+          {
+            event = ["LspAttach"];
+            desc = "LSP on-attach add keybinds [omnisharp-extended-lsp-nvim]";
+            callback =
+              mkLuaInline
+              /*
+              lua
```
snoweuph: @NotAShelf doesn't like these comments (he still needs to add the injections so we don't always smugle them in 👿 )

modules/plugins/languages/csharp.nix
```diff
@@ -288,20 +153,112 @@ in {
...
-        {
-          startPlugins = ["roslyn-nvim"];
-        }
-      ];
+    (mkIf (cfg.lsp.enable
+      && cfg.extensions.csharpls-extended-lsp-nvim.enable
+      && (elem "csharp_ls" cfg.lsp.servers)) {
+      vim = {
+        startPlugins = ["csharpls-extended-lsp-nvim"];
+        autocmds = [
```
snoweuph: @NotAShelf, @horriblename I don't think we come around using auto command here, or does one of you have a suggestion?

modules/plugins/utility/mason/mason.nix
```diff
@@ -0,0 +1,19 @@
+{lib, ...}: let
+  inherit (lib.options) mkEnableOption mkOption;
+  inherit (lib.types) listOf package;
+in {
+  options.vim.utility.mason = {
+    enable = mkEnableOption ''
+      a mock implementation of Mason.
+
+      ::: {.note}
+      This is intended to support plugins that rely on resolving asset paths within the Mason directory structure.
```
snoweuph: @NotAShelf @horriblename how do you guys feel about this?

snoweuph:
did a quick review, though I didn't look deeper into the lua code

modules/plugins/languages/csharp.nix
```diff
@@ -288,20 +153,112 @@ in {
...
+              value = formats.${name};
+            })
+            (filter (name: name != "lsp") cfg.format.type);
+        };
+      };
+    })
+
+    (mkIf (cfg.lsp.enable
+      && cfg.extensions.roslyn-nvim.enable
+      && (elem "roslyn-ls" cfg.lsp.servers)) {
```
snoweuph: if `"roslyn-ls"` is not in the `cfg.lsp.servers` but `cfg.extensions.roslyn-nvim.enable = true`, you should do an assert. The doc comment above, might not be seen, while a nix assert will be seen by anybody doing it wrong.

This also goes for the other LSP + plugins extensions

snoweuph:
@horriblename what would you think, about moving the plugins, that Only work with a specifc LSP, into the LSP preset, like

```nix
vim.lsp.presets.roslyn-ls = {
  enable = true;
  extensions.roslyn-nvim = {
    enable = true;
    setupOpts = {};
  };
};
``` 

This would make it clear from the Config structure, that the extensions doesn't work without the LSP. For Backwards compatibility and convenience, we could still have the options in the lang module, but just reference over to the preset.

modules/plugins/languages/csharp.nix
CaueAnjos: I can use `on_attach` instead of `autocmd` if you prefer. I tested it and it works equally well.

modules/plugins/utility/mason/mason.nix
CaueAnjos: There is a hacky solution to configure `roslyn-nvim` without this mock mason implementation (see https://github.com/NixOS/nixpkgs/issues/472357). But I personally don't like it, as it tends to break the output of the `:checkhealth roslyn` command, showing that everything is broken, even if it's working. This can make troubleshooting difficult in the future, since the user would have a misdiagnosis of the plugin.

Therefore, this is the best way to configure it without causing problems. But, if the idea of ​​a mock mason is bad, I could rewrite this extension to use the hacky solution. But, I really don't like it.

modules/plugins/languages/csharp.nix
snoweuph: please do. we try to minimize autocmds as much as possible, and `on_attach` is semanticly more correct.

modules/plugins/languages/csharp.nix
snoweuph: btw, main branch has now injections for mkLuaInline :3 just added them in

modules/plugins/utility/mason/mason.nix
horriblename: > it tends to break the output of the :checkhealth roslyn command

that sounds like something that should be fixed in roslyn-nvim

as far as I can see, the health check uses mason to look for the roslyn executable, the razor plugin, and a html-lsp (why?)

only the roslyn one is a hard error, but should be trivial to fix by check `vim.lsp.config.roslyn.cmd[1]` in https://github.com/seblyng/roslyn.nvim/blob/6a5e60a7c25d9ce0835aa9c69379f1c92e0a9d56/lua/roslyn/health.lua#L17

The other two may be a bit harder. the razor one can probably be checked just by parsing the flags in `vim.lsp.config.roslyn.cmd`, I'm not sure how the html lsp is supplied to roslyn-nvim so idk about that

in any case, I'd suggest making a issue/PR upstream

CaueAnjos:
Since this pull request was merged (seblyng/roslyn.nvim#292), there is a simple way to change the Razor extension path and pass arguments directly, without relying on a mock version of Mason.

snoweuph:
Your tree contains the commits for adding the mock mason, and alot of the asked changes, and thus has grown. Would you mind compressing it a bit more in like, i dunno 4 to 8 commits? after all git pulls all changes...

modules/plugins/lsp/presets/omnisharp.nix
```diff
@@ -0,0 +1,98 @@
...
+            )
+          end
+      '';
+      init_options = {};
+      capabilities = {
+        workspace = {
+          workspaceFolders = false; # https://github.com/OmniSharp/omnisharp-roslyn/issues/909
+        };
+      };
+      settings = {
```
snoweuph, line 51: @horriblename the whole Idea of the presets was to NOT have to much settings and docs for them in the mods, and only add changes ontop of defaults to make the behave correctly and not list all options.

how would you prefer this -> drop all things that are `default = null` OR add a `preset.<name>.settings` to this one and allow adding opts with docs in presets? I know you didn't like the later, last time I checked.
There prolly is some page of the LSP that lists all available settings, so that link could just be added into the enable options description as a user reference when doing the first one.

horriblename:
> @horriblename what would you think, about moving the plugins, that Only work with a specifc LSP, into the LSP preset, like
> 
> ```nix
> vim.lsp.presets.roslyn-ls = {
>   enable = true;
>   extensions.roslyn-nvim = {
>     enable = true;
>     setupOpts = {};
>   };
> };
> ``` 
> 
> This would make it clear from the Config structure, that the extensions doesn't work without the LSP. For Backwards compatibility and convenience, we could still have the options in the lang module, but just reference over to the preset.


I don't think there are any problems with the current structure so im leaning towards not moving it. we can discuss this in a new issue but it's not necessary for this pr

modules/plugins/lsp/presets/omnisharp.nix
snoweuph: nah fuck it, lets just keep it like it is RN, and do deeper discussions after this is merged

snoweuph:
@CaueAnjos when youre ready for the next review, just press this:
image

snoweuph:
LGTM, but want a final look from one of the others.
@horriblename @Soliprem @NotAShelf

modules/plugins/lsp/presets/csharp_ls.nix
```diff
@@ -0,0 +1,51 @@
...
+        function(dispatchers, config)
+           return vim.lsp.rpc.start({ '${getExe pkgs.csharp-ls}', '--features', 'razor-support' , '--features', 'metadata-uris'}, dispatchers, {
+            -- csharp-ls attempt to locate sln, slnx or csproj files from cwd, so set cwd to root directory.
+            -- If cmd_cwd is provided, use it instead.
+            cwd = config.cmd_cwd or config.root_dir,
+            env = config.cmd_env,
+            detached = config.detached,
+          })
+        end
+      '';
```
horriblename, line 31: wouldn't something like https://github.com/neovim/nvim-lspconfig/pull/4166/changes make more sense?

horriblename (changes requested):

modules/plugins/lsp/presets/csharp_ls.nix
CaueAnjos: This is outdated. `csharp-ls` now uses a custom RPC initialization, and relying solely on `cmd_cwd` is no longer sufficient.

The current configuration is better because it explicitly controls the working directory while still utilizing `root_dir` for proper workspace management.

Also, why not use the `nvim-lspconfig` configuration?

horriblename: > Also, why not use the nvim-lspconfig configuration?

we don't want to rely on external plugins for basic lsp configurations. Maybe @NotAShelf has some other reasoning I can't recall. That said, the lsp configs we do have are getting more and more complex that I'm second-guessing our decision

snoweuph: most LSPs work perfectly as LSPs, using a plugin just because a few LSPs are just ass under the hood and behave extremly weird, shouldnt be a reason for using a plugin to configure them all.

that `chsarp-ls` uses some custom RPC is completly against the whole fucking idea of LSPs which is standardization.

@CaueAnjos which lsp for C# are you actively using?

horriblename (approved):
thanks!

PR #1517, opened 2026-04-13
package/prettier-svelte: fix missing deps

horriblename:
supercedes #1516

PR #1529, opened 2026-04-17
prettier-plugin-svelte: skip npm prune step; remove redundant build deps

NotAShelf:
Supersedes #1517


pyrox0 (approved):
as the resident nodejs hater i dont like it but it works, merge it oomf

PR #1545, opened 2026-04-24
languages/haskell.nix: fix invalid keys in the haskell-tools config

dathegreat:
The haskell language module passes several keys to `vim.g.haskell_tools` that are not recognised by haskell-tools.nvim, causing it to emit warnings on every Haskell buffer:

```
unrecognized configs in vim.g.haskell_tools: { "hls.filetypes", "hls.root_dir", "tools.hover.enable", "hls.enable" }
```

The root cause is that the `hls` attrset in the module was shaped like a generic lspconfig server definition rather than the `haskell-tools.lsp.ClientOpts` type that haskell-tools.nvim actually expects.

### The following keys have been removed or corrected:

- `hls.enable` is not a recognised field; replaced with `hls.auto_attach`, which is the correct `haskell-tools.lsp.ClientOpts` field for controlling whether the LSP client attaches automatically.
- `hls.filetypes` is not a recognised field; haskell-tools.nvim manages filetype association internally and does not accept this key.
- `hls.root_dir` is not a recognised field; haskell-tools.nvim handles project root detection internally via `hie.yaml`, `stack.yaml`, `cabal.project`, etc.
- `tools.hover.enable` is not a recognised field; the `haskell-tools.hover.Opts` type only exposes `border`, `stylize_markdown`, and `auto_focus`.

### Additional improvements included in this fix:

- `hls.settings` is changed from a plain table to a function calling `ht.lsp.load_hls_settings(project_root)`, restoring per-project `hls.json` discovery. Explicit overrides are moved to `hls.default_settings`, which is the intended fallback mechanism.
- `ht.repl.toggle` in the `on_attach` keymaps is replaced with `vim.cmd('Haskell repl toggle')`, as `ht.repl.toggle` is not part of the documented public Lua API.
- The DAP binary path is resolved with `getExe'` rather than manual `/bin/` string interpolation, consistent with how the HLS binary is already resolved.
- `cabal-fmt` is added to the environment so the declared `cabalFormattingProvider = "cabal-fmt"` has a binary to call.
- The default `formattingProvider` is changed from `"ormolu"` to `"fourmolu"` to match the haskell-tools.nvim upstream default. Fourmolu is a strict superset of ormolu and is the more common choice in modern Haskell projects.

## Sanity Checking

- Style and consistency
- Package(s) built:
  - [ ] `.#maximal`
- Tested on platform(s)
  - [x] `x86_64-linux`
  - [ ] `aarch64-linux`
  - [ ] `x86_64-darwin`
  - [ ] `aarch64-darwin`

---


NotAShelf:
Was this PR guided or/and authored by a LLM?

dathegreat:
@mrcjkb Thank you for creating the wonderful haskell-tools! I'm pinging you here in case you have time to review how haskell-tools is being integrated into NVF. I would love your two cents-- especially if I am doing something wrong or sub-optimally.

dathegreat:
> Was this PR guided or/and authored by a LLM?

Claude aided me in troubleshooting and drafting the changes. You could say it was co-authored by an LLM, I suppose.

NotAShelf:
The prose in the issue body is very AI-like. Not that it's a problem per se, but I'd appreciate a disclosure next time.

NotAShelf (changes requested):
Diff looks good overall, but I must ask you to revert the formatting changes. Those usually happen when you auto-format with something like `nixfmt` and then run `nix fmt` for a second pass with alejandra.

dathegreat:
> The prose in the issue body is very AI-like. Not that it's a problem per se, but I'd appreciate a disclosure next time.

Of course! I apologize for the AI jump-scare. 

> Diff looks good overall, but I must ask you to revert the formatting changes. Those usually happen when you auto-format with something like nixfmt and then run nix fmt for a second pass with alejandra.

Done. This is also good to know for next time, thank you.

snoweuph:
I feel like similar to #1515 `hls` should be moved to `lsp/presets` and the haskell tools should be moved into `languages.haskell.extensions` simmilar to rust and java lang mods.

but I guess that is a topic for after this MR :thinking: 
[@I'm scared my haskell toolchain will break if I updated my system](https://github.com/horriblename): your up for the task? xD

docs/manual/release-notes/rl-0.9.md
```diff
@@ -185,6 +185,13 @@
 - Updated nix language plugin to use pkgs.nixfmt instead of
   pkgs.nixfmt-rfc-style
 
+[dathegreat](https://github.com/dathegreat):
+
+- Fixed invalid keys in the haskell-tools configuration
+- Changed the default haskell formatter to
```
snoweuph: what was it before? is it configurable?
changing defaults is a breaking change.

dathegreat: That's my bad. I thought `fourmolu`'s default behavior was identical to `ormolu`, but I was wrong. I set the default back to `ormolu` and made the formatter configurable.

dathegreat:
> I feel like similar to #1515 `hls` should be moved to `lsp/presets` and the haskell tools should be moved into `languages.haskell.extensions` simmilar to rust and java lang mods.
> 
> but I guess that is a topic for after this MR 🤔 [@I'm scared my haskell toolchain will break if I updated my system](https://github.com/horriblename): your up for the task? xD

I took a swing at splitting the configs up. 

Also, `@I'm scared my haskell toolchain will break if I updated my system` made me laugh 😂 I remember reading their comment about toolchain anxiety a while back. Very relatable.

modules/plugins/lsp/presets/hls.nix
```diff
@@ -0,0 +1,42 @@
...
+  cfg = config.vim.lsp.presets.hls;
+in {
+  options.vim.lsp.presets.hls = {
+    enable = mkLspPresetEnableOption "hls" "Haskell" [];
+  };
+
+  config = mkIf cfg.enable {
+    vim.lsp.servers.hls = {
+      enable = true;
+      cmd = [(getExe' pkgs.haskellPackages.haskell-language-server "haskell-language-server-wrapper") "--lsp"];
```
snoweuph: the preset should be renamed to `haskell-language-server` to match the rest of our preset naming, which follows the nixpkgs.

modules/plugins/lsp/presets/hls.nix
```diff
@@ -0,0 +1,42 @@
...
+      filetypes = ["haskell" "lhaskell"];
+      root_dir =
+        mkLuaInline
+        /*
+        lua
```
snoweuph: please remove this. If you use the latest nvf version from main, you will have injections without the comment

modules/plugins/lsp/presets/hls.nix
```diff
@@ -0,0 +1,42 @@
...
+in {
+  options.vim.lsp.presets.hls = {
+    enable = mkLspPresetEnableOption "hls" "Haskell" [];
+  };
+
+  config = mkIf cfg.enable {
+    vim.lsp.servers.hls = {
+      enable = true;
+      cmd = [(getExe' pkgs.haskellPackages.haskell-language-server "haskell-language-server-wrapper") "--lsp"];
+      filetypes = ["haskell" "lhaskell"];
```
snoweuph: presets shouldn't contain the filetypes, filetypes are assigned from the language modules, when enabeling the LSPs. this exists because there are quite a lot of LSPs that are used across language modules.

modules/plugins/lsp/presets/hls.nix
```diff
@@ -0,0 +1,42 @@
...
+        */
+        ''
+          function(bufnr, on_dir)
+            local fname = vim.api.nvim_buf_get_name(bufnr)
+            on_dir(util.root_pattern('hie.yaml', 'stack.yaml', 'cabal.project', '*.cabal', 'package.yaml')(fname))
+          end
+        '';
+      settings = {
+        haskell = {
+          formattingProvider = mkDefault "ormolu";
```
snoweuph: why the excplicit `mkDefault` instead of direct asignment?

modules/plugins/lsp/presets/hls.nix
```diff
@@ -0,0 +1,42 @@
...
+      filetypes = ["haskell" "lhaskell"];
+      root_dir =
+        mkLuaInline
+        /*
+        lua
+        */
+        ''
+          function(bufnr, on_dir)
+            local fname = vim.api.nvim_buf_get_name(bufnr)
+            on_dir(util.root_pattern('hie.yaml', 'stack.yaml', 'cabal.project', '*.cabal', 'package.yaml')(fname))
```
snoweuph: wouldn't `root_makers = ["hie.yaml" "stack.yaml" "stack.yml" "cabal.project"];` be enough? why do explicitly use a lua function?

modules/plugins/lsp/presets/hls.nix
```diff
@@ -0,0 +1,42 @@
...
+  lib,
+  pkgs,
+  ...
+}: let
+  inherit (lib.meta) getExe';
+  inherit (lib.modules) mkDefault mkIf;
+  inherit (lib.nvim.types) mkLspPresetEnableOption;
+  inherit (lib.generators) mkLuaInline;
+
+  cfg = config.vim.lsp.presets.hls;
```
snoweuph: you are never enableing the preset in the haskell language module.

snoweuph (changes requested):
Needs some work on the preset.

see:
- <https://github.com/NotAShelf/nvf/blob/main/modules/plugins/languages/wgsl.nix>
- <https://github.com/NotAShelf/nvf/blob/main/modules/plugins/lsp/presets/wgsl-analyzer.nix>

for simple example how lang modules and lsp presets work together

modules/plugins/lsp/presets/hls.nix
snoweuph: if we can do without extra lua, we try to do it without extra lua, especially as the neovim builtin LSP support allready is pretty good.

snoweuph:
> I took a swing at splitting the configs up.

nice, that you are taking the time :3


> Also, `@I'm scared my haskell toolchain will break if I updated my system` made me laugh 😂 I remember reading their comment about toolchain anxiety a while back. Very relatable.

xD

modules/plugins/lsp/presets/hls.nix
dathegreat: The preset is intentionally disabled if `haskell-tools` is enabled. I added a comment explaining this reasoning. Does it make sense now, or am I still not enabling properly?

modules/plugins/lsp/presets/hls.nix
dathegreat: Mostly habit. If something is meant to be overridden, I figure it should have a low priority. You're right though, in this case it doesn't matter.

NotAShelf:
CC @horriblename as the resident Haskell enjoyer

horriblename (approved):

horriblename (dismissed):

NotAShelf (dismissed):

dathegreat:
> Needs some work on the preset.
> 
> see:
> 
>     * https://github.com/NotAShelf/nvf/blob/main/modules/plugins/languages/wgsl.nix
> 
>     * https://github.com/NotAShelf/nvf/blob/main/modules/plugins/lsp/presets/wgsl-analyzer.nix
> 
> 
> for simple example how lang modules and lsp presets work together

@snoweuph Fixed! Please let me know if you had something different in mind.

NotAShelf:
Wonder why auto-merge, well, auto-merged...

NotAShelf:
Ah, missing one review nvm. @snoweuph can merge this when approved.

snoweuph:
Gimme till weekend, RL stressy RN.

snoweuph:
gonna review RN :3

modules/plugins/languages/haskell.nix
```diff
@@ -4,63 +4,23 @@
...
+  inherit (lib.nvim.dag) entryAfter;
   inherit (lib.meta) getExe';
-  inherit (lib.generators) mkLuaInline;
   inherit (pkgs) haskellPackages;
 
   cfg = config.vim.languages.haskell;
 
-  defaultServers = ["hls"];
+  defaultServers = ["haskell-language-server"];
   servers = {
```
snoweuph: Should be a list of strings to match all other modules

```suggestion
  servers = ["haskell-language-server"];
```

modules/plugins/lsp/presets/haskell-language-server.nix
```diff
@@ -0,0 +1,31 @@
...
+  inherit (lib.nvim.types) mkLspPresetEnableOption;
+
+  cfg = config.vim.lsp.presets.haskell-language-server;
+in {
+  options.vim.lsp.presets.haskell-language-server = {
+    enable = mkLspPresetEnableOption "haskell-language-server" "Haskell" [];
+  };
+
+  config = mkIf cfg.enable {
+    vim.extraPackages = [pkgs.haskellPackages.cabal-fmt];
```
snoweuph: prefferebly dont add them to `vim.extraPackage` but wrap them with the lsp package.

We do a simmilar thing with uncrustify for the vala lsp.

https://github.com/NotAShelf/nvf/blob/main/modules/plugins/lsp/presets/vala-language-server.nix#L24


alternatively, if `settings.haskell.cabalFormattingProvider` supports paths do something like

`cabalFormattingProvider = getExe pkgs.haskellPackages.cabal-fmt`

modules/plugins/lsp/presets/haskell-language-server.nix
```diff
@@ -0,0 +1,31 @@
...
+    vim.lsp.servers.haskell-language-server = {
+      enable = true;
+      cmd = [(getExe' pkgs.haskellPackages.haskell-language-server "haskell-language-server-wrapper") "--lsp"];
+      root_markers = ["hie.yaml" "stack.yaml" "cabal.project" "*.cabal" "package.yaml"];
+      settings = {
+        haskell = {
+          formattingProvider = "ormolu";
```
snoweuph: what is ormolu?
does it need to be provides by something?
Might we need to wrap an extra package for it?

Add a comment explaining

modules/plugins/languages/haskell.nix
```diff
@@ -113,34 +84,67 @@ in {
       };
     })
 
-    (mkIf (cfg.dap.enable || cfg.lsp.enable) {
+    # haskell-tools prefers to manage the lsp directly,
+    # so we only configure the lsp ourselves if haskell-tools is disabled
+    (mkIf (cfg.lsp.enable && !cfg.extensions.haskell-tools.enable) {
```
snoweuph: Do only this here:
```suggestion
    (mkIf cfg.lsp.enable) {
```

Instead assert that the following is never true.
```nix
cfg.lsp.enable && cfg.extensions.haskell-tools.enable
```
If it is true, we want to clearly assert and give your comment to the user of nvf. The Users wont read comments in the code, so that way, they actually get the info, instead of possibly unexpected behaviour to them.

modules/plugins/languages/haskell.nix
```diff
@@ -113,34 +84,67 @@ in {
...
+              settings = {
+                haskell = {
+                  formattingProvider = "${cfg.lsp.formattingProvider}",
+                  cabalFormattingProvider = "cabal-fmt",
                 },
               },
-              hls = ${toLuaObject servers.hls},
-            ''}
+            },
             ${optionalString cfg.dap.enable ''
```
snoweuph: Why is the dap also in Lua, and only enabled with haskell-tools together?
doesnt it work standalone?

modules/plugins/languages/haskell.nix
```diff
@@ -88,6 +48,11 @@ in {
         default = defaultServers;
         description = "Haskell LSP server to use";
       };
+      formattingProvider = mkOption {
+        type = enum ["ormolu" "fourmolu" "stylish-haskell" "brittany" "floskell" "none"];
```
snoweuph: Where do these formatter come from?
Normally we have a section of the formatters at the top like:
https://github.com/NotAShelf/nvf/blob/main/modules/plugins/languages/python.nix#L24

and use conform-nvim for configuration:
https://github.com/NotAShelf/nvf/blob/main/modules/plugins/languages/nix.nix#L207

snoweuph (changes requested):
Looks pretty good already, but some nicks.

modules/plugins/languages/haskell.nix
horriblename: we rely on haskell-tools to provide DAP config

we do that for other languages as well, since DAP configs are usually a lot more complicated than LSP (at least one that is usable)

horriblename:
Ok, there are more problems with haskell-tools than the ones listed in this PR. I'll migrate it to the extensions namespace + do setupOpts refactor and see what else needs to be done

dathegreat:
@snoweuph thanks for another helpful review! I believe I have addressed everything you pointed out.

@horriblename I believe I already migrated haskell-tools to the extension namespace. What else needs to be done?

modules/plugins/languages/haskell.nix
```diff
@@ -103,44 +82,128 @@ in {
...
+              '';
+            };
+            on_attach = mkOption {
+              type = nullOr luaInline;
+              description = "Function to run when HLS is attached";
+              default = mkLuaInline ''
+                function(client, bufnr)
+                  local ht = require("haskell-tools")
+                  local opts = { noremap = true, silent = true, buffer = bufnr }
+                  vim.keymap.set('n', '<localleader>cl', vim.lsp.codelens.run, opts)
```
snoweuph: These should prolly be factored into some config opts.

something like `extensions.haskell-tools.keymaps.*`

snoweuph (dismissed):
LGTM

modules/plugins/lsp/presets/haskell-language-server.nix
```diff
@@ -0,0 +1,44 @@
...
+              --prefix PATH : ${pkgs.haskellPackages.cabal-fmt}/bin
+          '';
+        }) "haskell-language-server-wrapper")
+        "--lsp"
+      ];
+      root_markers = ["hie.yaml" "stack.yaml" "cabal.project" "*.cabal" "package.yaml"];
+      settings = {
+        haskell = {
+          # formatting is handled by conform-nvim; disable HLS's built-in formatter
+          formattingProvider = "none";
```
horriblename:
```suggestion
```

don't change these settings, if conform can find a formatter it won't touch the LSP, there is no need to change this

horriblename (changes requested):

modules/plugins/languages/haskell.nix
dathegreat: Done! I used `mappings` instead of `keymaps` to match existing patterns.

modules/plugins/languages/haskell.nix
```diff
@@ -103,44 +86,152 @@ in {
...
+            else generatedOnAttach;
+        in
+          htCfg.setupOpts
+          // {
+            hls =
+              htCfg.setupOpts.hls
+              // {
+                on_attach = effectiveOnAttach;
+              };
+          };
```
horriblename:
```suggestion
        globals.haskell_tools = cfg.extensions.haskell-tools.setupOpts;
        languages.haskell.extensions.haskell_tools.setupOpts.on_attach = let
          htCfg = cfg.extensions.haskell-tools;
          keymapDefinitions = options.vim.languages.haskell.extensions.haskell-tools.mappings;
          mappings = addDescriptionsToMappings htCfg.mappings keymapDefinitions;
          mkBinding = binding: action:
            if binding.value != null
            then "vim.keymap.set('n', ${toLuaObject binding.value}, ${action}, {buffer=bufnr, noremap=true, silent=true, desc=${toLuaObject binding.description}})"
            else "";
        in mkLuaInline ''
            function(client, bufnr)
              local ht = require("haskell-tools")
              ${mkBinding mappings.codeLensRun "vim.lsp.codelens.run"}
              ${mkBinding mappings.hoogleSignature "ht.hoogle.hoogle_signature"}
              ${mkBinding mappings.evalAll "ht.lsp.buf_eval_all"}
              ${mkBinding mappings.replToggle "function() vim.cmd('Haskell repl toggle') end"}
              ${mkBinding mappings.replToggleFile "function() vim.cmd('Haskell repl toggle ' .. vim.api.nvim_buf_get_name(0)) end"}
              ${mkBinding mappings.replQuit "function() vim.cmd('Haskell repl quit') end"}
            end
         '';

```

horriblename (changes requested):

horriblename (dismissed):
Thanks!

horriblename (dismissed):

horriblename:
if @snoweuph has no further comments we can merge

docs/manual/release-notes/rl-0.9.md
```diff
@@ -39,6 +39,9 @@
   `setupOpts.strategies.chat.adapter`, rename them to
   `setupOpts.interactions.chat.adapter`.
 
+- Haskell LSP now defaults to haskell-language-server, haskell-tools based LSP
```
snoweuph: Should be under the section of your username, not in the global section for brraking changes

modules/plugins/languages/haskell.nix
```diff
@@ -84,12 +53,26 @@ in {
...
 
+    format = {
+      enable =
+        mkEnableOption "Haskell formatting"
+        // {
+          default = config.vim.languages.enableFormat;
+          defaultText = literalExpression "config.vim.languages.enableFormat";
+        };
+      type = mkOption {
+        type = deprecatedSingleOrListOf "vim.languages.haskell.format.type" (enum (attrNames formats));
```
snoweuph: Dont use the `deprecatedSingleOrListOf`, when newly adding formatters, use `listOf`

modules/plugins/lsp/presets/haskell-language-server.nix
```diff
@@ -0,0 +1,42 @@
...
+      enable = true;
+      cmd = [
+        (getExe' (pkgs.symlinkJoin {
+          name = "haskell-language-server-wrapper";
+          paths = [pkgs.haskellPackages.haskell-language-server];
+          meta.mainProgram = "haskell-language-server-wrapper";
+          buildInputs = [pkgs.makeBinaryWrapper];
+          postBuild = ''
+            wrapProgram $out/bin/haskell-language-server-wrapper \
+              --prefix PATH : ${pkgs.haskellPackages.cabal-fmt}/bin
```
snoweuph: Add a comment adding why we are wrapping with `cabal-fmt`.

modules/plugins/lsp/presets/haskell-language-server.nix
```diff
@@ -0,0 +1,42 @@
...
+          '';
+        }) "haskell-language-server-wrapper")
+        "--lsp"
+      ];
+      root_markers = ["hie.yaml" "stack.yaml" "cabal.project" "*.cabal" "package.yaml"];
+      settings = {
+        haskell = {
+          # cabal-fmt is an external tool; it is wrapped into the LSP binary's PATH above
+          cabalFormattingProvider = "cabal-fmt";
```
snoweuph: Couldnt this be `cabalFormattingProvider = getExe pkgs.haskellPackages.cabal-fmt;` instead of wrapping? That would also allow NVF users to override this one if they want, while doing so would be alot more anoying with the wrapper LSP package

snoweuph (changes requested):
A few small nicks left, after those its good to me

docs/manual/release-notes/rl-0.9.md
horriblename: did I stutter

(that is indeed a notice of breaking change, I added that)

snoweuph: There are username sections in the breaking and non breaking changes. Shouldn't it be moved in your named section in the breaking section?

modules/plugins/languages/haskell.nix
```diff
@@ -88,6 +86,59 @@ in {
     extensions = {
       haskell-tools = {
         enable = mkEnableOption "haskell-tools.nvim";
+
+        setupOpts = mkPluginSetupOption "haskell-tools.nvim" {
+          hls = {
+            cmd = mkOption {
+              type = nullOr (listOf str);
+              default = [
+                "${pkgs.haskellPackages.haskell-language-server}/bin/haskell-language-server"
```
dathegreat: @horriblename I'm curious why you replaced `haskell-language-server-wrapper` with `haskell-language-server`.

> Since HLS is very sensitive about the GHC version, there's a
> very high chance that the default HLS we use is not
> compatible with your project. It is highly recommended to set
> this option to `null` and install HLS separately in a
> `devShell`.

^ Your note is only true for the bare `haskell-language-server`. The wrapper is much more resilient to different GHC versions. From the [nixpkgs haskell docs](https://github.com/NixOS/nixpkgs/blob/85676b222a7a18e2ee1e64a2a916201abe2b8889/doc/languages-frameworks/haskell.section.md?plain=1#L759):

> When you run `haskell-language-server-wrapper` it will detect the GHC
> version used by the project you are working on (by asking e.g. cabal or
> stack) and pick the appropriate versioned binary from your path.

And in fact the docs recommend only using `pkgs.haskell-language-server` in something like a nix-shell environment: 
 
> The top level `pkgs.haskell-language-server` attribute is just a convenience
> wrapper to make it possible to install HLS for multiple GHC versions at the
> same time. If you know, that you only use one GHC version, e.g., in a project
> specific `nix-shell` you can use
> `pkgs.haskellPackages.haskell-language-server` or
> `pkgs.haskell.packages.*.haskell-language-server` from the package set you use.

horriblename: from my testing, `-wrapper` can only pick up hls from $PATH, it does not look at the non-wrapper one in the same package (unless its in PATH ofc)

_maybe_ adding hls to PATH would solve this, but there is also the fact that, even if hls matches your ghc major+minor version, it might still not work cuz it's not the ghc hls is built on (usually due to different nixpkgs versions of nvf and devShell). Maybe the wrapper is smart enough to resolve this itself? I haven't tested.

modules/plugins/lsp/presets/haskell-language-server.nix
horriblename: it is apparently possible to set the location of cabal-fmt with something like this:

```
haskell.plugin.cabal-fmt.path =... 
```

I couldn't find clear docs on this so the exact key might differ slightly, but the flag is indeed defined in the src code of the plugin so it _has_ to be configurable somehow right?

docs/manual/release-notes/rl-0.9.md
horriblename: @NotAShelf uhm, why is there an "unnamed" section and named ones in #breaking changes?

modules/plugins/languages/haskell.nix
dathegreat: I have been using the wrapper every day at my job, and it is smart enough to "just work" in all my projects. I did not even realize it was doing so much heavy lifting until I pulled your changes and then all my projects broke 😂. You should give it a try, it's pretty amazing.

modules/plugins/lsp/presets/haskell-language-server.nix
dathegreat: You are my hero for finding that config option, @horriblename ! I unwrapped the package and added an option for users to configure either `cabal-fmt` or `cabal-gild` as the formatter.

docs/manual/release-notes/rl-0.9.md
dathegreat: > @NotAShelf uhm, why is there an "unnamed" section and named ones in #breaking changes?

Maybe someone's name is "unnamed" 🤔

modules/plugins/lsp/presets/haskell-language-server.nix
```diff
@@ -0,0 +1,45 @@
...
+        "--lsp"
+      ];
+      root_markers = ["hie.yaml" "stack.yaml" "cabal.project" "*.cabal" "package.yaml"];
+      settings = mkIf cabalFmtCfg.enable {
+        haskell = {
+          cabalFormattingProvider = cabalFmtCfg.type;
+          # This option is undocumented in the haskell-language-server docs, but it does exist in the plugin API
+          # https://github.com/haskell/haskell-language-server/blob/f158128ec034bec1667c440262641086bbb4d359/plugins/hls-cabal-fmt-plugin/src/Ide/Plugin/CabalFmt.hs#L51
+          plugin.${cabalFmtCfg.type}.config.path = getExe cabalFormats.${cabalFmtCfg.type};
+        };
```
horriblename: I think formatting should just be handled by conform. This hls config in particular doesn't even attach to cabal files. And if you try to attach to cabal files, it does not work well with lsp-signature (spams errors)

The haskell-tools lsp still attaches to cabal files, and spams errors if lsp-signature is enabled, but there's not much we can do there (I lied, there are but it's annoying as hell to do so) so just add a warning when both haskell-tools and lsp-signature is enabled for now. I'll try to come up with something less bad tomorrow

modules/plugins/languages/haskell.nix
```diff
@@ -103,44 +106,157 @@ in {
...
+                  ${mkBinding mappings.replToggleFile "function() vim.cmd('Haskell repl toggle ' .. vim.api.nvim_buf_get_name(0)) end"}
+                  ${mkBinding mappings.replQuit "function() vim.cmd('Haskell repl quit') end"}
+                end
+              '';
+            settings = mkIf cfg.cabalFormat.enable (mkDefault {
+              haskell = {
+                cabalFormattingProvider = cfg.cabalFormat.type;
+                plugin.${cfg.cabalFormat.type}.config.path = getExe cabalFormats.${cfg.cabalFormat.type};
+              };
+            });
```
horriblename: handle cabal formatting in conform (see my other comment below)

modules/plugins/languages/haskell.nix
```diff
@@ -103,44 +106,157 @@ in {
...
+              htCfg = cfg.extensions.haskell-tools;
+              keymapDefinitions = options.vim.languages.haskell.extensions.haskell-tools.mappings;
+              mappings = addDescriptionsToMappings htCfg.mappings keymapDefinitions;
+              mkBinding = binding: action:
+                if binding.value != null
+                then "vim.keymap.set('n', ${toLuaObject binding.value}, ${action}, {buffer=bufnr, noremap=true, silent=true, desc=${toLuaObject binding.description}})"
+                else "";
+            in
+              mkLuaInline ''
+                function(client, bufnr)
```
horriblename: note to self: is it possible to mask client/server cap for textDocument/signatureHelp here? or disable lsp-signature somehow

modules/plugins/lsp/presets/haskell-language-server.nix
```diff
@@ -0,0 +1,45 @@
...
+
+  config = mkIf cfg.enable {
+    vim.lsp.servers.haskell-language-server = {
+      enable = true;
+      cmd = [
+        (getExe' (pkgs.symlinkJoin {
+          name = "haskell-language-server-wrapper";
+          paths = [pkgs.haskellPackages.haskell-language-server];
+          meta.mainProgram = "haskell-language-server-wrapper";
+          buildInputs = [pkgs.makeBinaryWrapper];
```
horriblename, line 25:
```suggestion
          buildInputs = [pkgs.makeBinaryWrapper];
          postBuild = ''
            # wrap HLS-wrapper so it can find the actual binary
            wrapProgram $out/bin/haskell-language-server-wrapper \
              --suffix PATH : ${pkgs.haskellPackages.haskell-language-server}/bin
          '';
```

horriblename: same thing should be done on the haskell-tools side.

modules/plugins/languages/haskell.nix
```diff
@@ -103,44 +106,157 @@ in {
...
+          replToggleFile = mkMappingOption "Toggle REPL for current file [haskell-tools.nvim]" "<localleader>rf";
+          replQuit = mkMappingOption "Quit REPL [haskell-tools.nvim]" "<localleader>rq";
+        };
+
+        setupOpts = mkPluginSetupOption "haskell-tools.nvim" {
+          hls = {
+            cmd = mkOption {
+              type = nullOr (listOf str);
+              default = [
+                "${pkgs.haskellPackages.haskell-language-server}/bin/haskell-language-server-wrapper"
```
horriblename: wrap with itself (see https://github.com/NotAShelf/nvf/pull/1545/changes#r3253924623)

horriblename (changes requested):

modules/plugins/languages/haskell.nix
horriblename: yea, it does indeed work, as long as hls is in PATH

dathegreat:
@horriblename I have been daily driving your changes and everything looks great! Thank you for working with me on this.

It looks like the release-notes nit is the only requested change left, and I feel unqualified to take part in that debate. @snoweuph Is that section enough of a blocker to keep this PR from merging?

docs/manual/release-notes/rl-0.9.md
snoweuph: > @NotAShelf uhm, why is there an "unnamed" section and named ones in #breaking changes?

Just create a named section with your name.

snoweuph:
> @horriblename I have been daily driving your changes and everything looks great! Thank you for working with me on this.
> 
> It looks like the release-notes nit is the only requested change left, and I feel unqualified to take part in that debate. @snoweuph Is that section enough of a blocker to keep this PR from merging?

Changelogs are important. As this doesnt do any breaking changes, I would be okay, with droping that part, but I know that Raf likes his changelogs.

Also the one to do the final decision on this one beeing ready has to be @horriblename .

horriblename (approved):

horriblename:
@snoweuph do you want to review it one final time?

horriblename (dismissed):

snoweuph:
Yes, later today

modules/plugins/languages/haskell.nix
```diff
@@ -103,44 +97,159 @@ in {
...
-                cmd = ${
-                if isList cfg.dap.package
-                then toLuaObject cfg.dap.package
-                else ''{"${cfg.dap.package}/bin/haskell-debug-adapter"}''
-              },
-              },
-            ''}
-            }
-          '';
+        globals.haskell_tools = cfg.extensions.haskell-tools.setupOpts;
```
snoweuph, line 252: Does this have to configured via Globals?
Cant we go with `vim.lazy.plugins` for this one and use `setupOpts`?
or does the plugin not support that?

modules/plugins/lsp/lsp-signature/lsp-signature.nix
```diff
@@ -1,11 +1,40 @@
...
-  inherit (lib.nvim.types) mkPluginSetupOption;
+  inherit (lib.options) mkEnableOption mkOption;
+  inherit (lib.nvim.types) mkPluginSetupOption luaInline;
+  inherit (lib.generators) mkLuaInline;
 in {
   options.vim.lsp = {
     lspSignature = {
       enable = mkEnableOption "lsp signature viewer";
-      setupOpts = mkPluginSetupOption "lsp-signature" {};
+      setupOpts = mkPluginSetupOption "lsp-signature" {
```
snoweuph, line 9: This is new? I dont see anything about this in the Changelog

snoweuph (dismissed):
LGTM, besides the small 2 things

modules/plugins/languages/haskell.nix
horriblename: doesn't matter imo, the plugin lazy-loads itself, adding it to vim.lazy is just more work for nothing

snoweuph (approved):
LGTM

PR #1581, opened 2026-05-10
languages/clang: add formatting

### LINKED ISSUE #1564: Improve C Direct Formatting and via LSP
### I have verified that this feature request has not been made before

- [x] I have checked the [issues tab](https://github.com/notashelf/nvf/issues?q=is%3Aissue), and did not find a similar feature request. I understand that my issue will be closed if it is a duplicate.

### Feature Type

Other

### Feature description

Clangd sucks, when it comes to formatting. It ignores the configuration of the Editor.
The Solution is to wrap it with uncrustify, like we do for the vala module.

### Alternatives

add a flag to the clangd preset to only enable uncrustify if you want it.
like `vim.lsp.presets.clangd.extensions.uncrustify.enable`.

### Additional Context

- https://uncrustify.sourceforge.net/
- https://github.com/NotAShelf/nvf/blob/main/modules/plugins/lsp/presets/vala-language-server.nix#L29

snoweuph:
Adds formatters to C.

## Sanity Checking

- Style and consistency
- Package(s) built:
  - [x] `.#maximal`
- Tested on platform(s)
  - [x] `x86_64-linux`
  - [ ] `aarch64-linux`
  - [ ] `x86_64-darwin`
  - [ ] `aarch64-darwin`

---

### DIFF
--- a/modules/plugins/languages/clang.nix
+++ b/modules/plugins/languages/clang.nix
@@ -8,9 +8,12 @@
   inherit (lib.options) mkEnableOption mkOption literalExpression;
   inherit (lib.types) bool enum package listOf;
   inherit (lib) genAttrs;
+  inherit (lib.meta) getExe getExe';
   inherit (lib.modules) mkIf mkMerge;
   inherit (lib.nvim.types) mkGrammarOption;
+  inherit (lib.generators) mkLuaInline;
   inherit (lib.nvim.dag) entryAfter;
+  inherit (lib.nvim.attrsets) mapListToAttrs;
 
   cfg = config.vim.languages.clang;
 
@@ -44,6 +47,54 @@
         dap.configurations.c = dap.configurations.cpp
       '';
     };
+  };
+
+  defaultFormat = ["clang-format"];
+  formats = {
+    astyle = {
+      command = getExe pkgs.astyle;
+      stdin = false;
+      args = mkLuaInline ''
+        function(self, ctx)
+          local args = {
+            "$FILENAME",
+          }
+
+          if not vim.bo[ctx.buf].expandtab then
+            table.insert(args, "--indent=tab=" .. ctx.shiftwidth)
+          else
+            table.insert(args, "--indent=spaces=" .. ctx.shiftwidth)
+          end
+
+          return args
+        end
+      '';
+    };
+    indent = {
+      command = getExe pkgs.indent;
+      stdin = true;
+      args = mkLuaInline ''
+        function(self, ctx)
+          local args = {
+            "--indent-level", ctx.shiftwidth,
+            "--tab-size", ctx.shiftwidth,
+          }
+
+          if not vim.bo[ctx.buf].expandtab then
+            table.insert(args, "--use-tabs")
+          else
+            table.insert(args, "--no-tabs")
+          end
+
+          return args
+        end
+      '';
+      # Default is GNU style. Nobody likes that one.
+      # This is under `append_args`, to allow easy editing of this argument,
+      # without having to redefine everything as a user.
+      append_args = ["--linux-style"];
+    };
+    clang-format.command = getExe' pkgs.clang-tools "clang-format";
   };
 in {
   options.vim.languages.clang = {
@@ -102,6 +153,21 @@ in {
         default = debuggers.${cfg.dap.debugger}.package;
       };
     };
+
+    format = {
+      enable =
+        mkEnableOption "C formatting"
+        // {
+          default = config.vim.languages.enableFormat;
+          defaultText = literalExpression "config.vim.languages.enableFormat";
+        };
+
+      type = mkOption {
+        type = listOf (enum (attrNames formats));
+        default = defaultFormat;
+        description = "C formatter to use";
+      };
+    };
   };
 
   config = mkIf cfg.enable (mkMerge [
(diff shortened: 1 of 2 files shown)


modules/plugins/languages/clang.nix
```diff
@@ -45,6 +47,52 @@
...
+          end
+
+          return args
+        end
+      '';
+    };
+    indent = {
+      command = getExe pkgs.indent;
+      stdin = true;
+      args = mkLuaInline ''
```
snoweuph, line 76: I'm not going to try upstreaming till this is merged: <https://github.com/stevearc/conform.nvim/pull/862>.

It currently doesn't look like its worth trying to upstream into conform, as I still got no response.

modules/plugins/languages/clang.nix
```diff
@@ -45,6 +47,52 @@
       '';
     };
   };
+
+  defaultFormat = ["astyle"];
+  formats = {
+    astyle = {
+      command = getExe pkgs.astyle;
+      stdin = false;
+      args = mkLuaInline ''
```
snoweuph, line 57: I'm not going to try upstreaming till this is merged: <https://github.com/stevearc/conform.nvim/pull/862>.

It currently doesn't look like its worth trying to upstream into conform, as I still got no response.

modules/plugins/languages/clang.nix
```diff
@@ -45,6 +47,52 @@
...
+        function(self, ctx)
+          local args = {
+            "--mode=c",
```
snoweuph: We might wanna split the `clang` module into `c` and `cpp`, to have different configs in both. or alternatively test what conform provides us in the `ctx`, so we could possibly make this smarter.

snoweuph: image

nevermind

modules/plugins/languages/clang.nix
```diff
@@ -45,6 +47,52 @@
...
+            "$FILENAME",
+          }
+
+          if not vim.bo[ctx.buf].expandtab then
+            table.insert(args, "--indent=tab=" .. ctx.shiftwidth)
```
horriblename, line 64: is ctx.shiftwidth provided by conform or is this a typo for `vim.bo`?

modules/plugins/languages/clang.nix
```diff
@@ -45,6 +47,52 @@
...
+          else
+            table.insert(args, "--indent=spaces=" .. ctx.shiftwidth)
```
horriblename, line 66: as above

modules/plugins/languages/clang.nix
horriblename: it is indeed provided (but why)

modules/plugins/languages/clang.nix
horriblename: should `--mode` be omitted then? to allow user customization if they wanna use it for other filetypes. though they can't do they till we refactor conform into presets as well

snoweuph: I would keep it, for Beeing explizit. 

As soon as we refactor out formatters into presets, we should keep these kind of cases in mind though

modules/plugins/languages/clang.nix
```diff
@@ -45,6 +47,52 @@
...
+        end
+      '';
+    };
+    indent = {
+      command = getExe pkgs.indent;
+      stdin = true;
+      args = mkLuaInline ''
+        function(self, ctx)
+          local args = {
+            "--linux-style", -- default is GNU style. Nobody likes that one
```
horriblename: I'm a bit on the fence on this one, if we omit this, the user gets to change the default easily, by adding to append_args. If we keep this in, we get better defaults, but now the user needs to copy the whole ass function just to change one flag...

one alternative is to add it as append_args, so the user only needs to lib.mkForce ["--their-style"]

not a blocker as is, but I'd want your opinion on this

snoweuph: Totally with you on this one, will move the `*-style` int append args.

snoweuph:
rebased

modules/plugins/languages/clang.nix
```diff
@@ -45,6 +47,54 @@
       '';
     };
   };
+
+  defaultFormat = ["astyle"];
```
horriblename: just from the fact that I've never heard of the other ones, I think clang-format should be the default

irrelevant to this pr, I still don't know what to do with langs where LSP formatting and conform both have the same exact formatter, skipping the conform one prolly makes sense since it's prolly slower? idk

horriblename (changes requested):
looks good overall, just the default, I think astyle is a weird choice

modules/plugins/languages/clang.nix
snoweuph: > irrelevant to this pr, I still don't know what to do with langs where LSP formatting and conform both have the same exact formatter, skipping the conform one prolly makes sense since it's prolly slower? idk

Users can just not enable them, while we still provide the capability if they don't want to use the LSP for some reason.

for example, somebody still wants to use the formatter of the LSP, but a different LSP, with the current state this is easily possible.

snoweuph: > just from the fact that I've never heard of the other ones, I think clang-format should be the default

`clang-format` doesn't have option to be informed about tabs vs spaces and how much of them. `asytle` has objectively the best formatting of them all, and will for a user with the lua I wrote, just work as expected and respect `.editorconfig`.

If you really want, I will set the default to `clang-format` even though its worse.

funfact: I did only know about `astyle` before this and not about `clang-format`, because I had used it for java xD

snoweuph:
rebased

snoweuph:
rebased

snoweuph:
rebased

snoweuph:
rebased

snoweuph:
rebased

snoweuph:
rebased

snoweuph:
rebased and changed default to `clang-format`

snoweuph:
rebased

modules/plugins/languages/clang.nix
horriblename: I looked around a couple C/++ projects that I know of (llama-cpp, hyprland, roc-toolkit, linux), and most seem to have a .clang-format file, I think that's a pretty good indication of what people would most likely want :)

horriblename (approved):

PR #1634, opened 2026-06-11
wrapper: expose `appName` in `mnw` via `config.vim.appname`

NotAShelf:
Fixes the hardcoded appname in the build wrapper. I've noticed that the way we set wrapper options is a bit... messy. We have `vim.build`, but it's used only for `finalPackage`. I'd like to establish a naming convention with @horriblename, @snoweuph and @Soliprem so this PR can be repurposed as a general-purpose wrapper refactoring one. Alternatively we should allow additional args to be passed to mnw. Maybe @Gerg-L can chime in on that. Recursive merge? Just `//`?

### DIFF
--- a/modules/wrapper/build/config.nix
+++ b/modules/wrapper/build/config.nix
@@ -80,7 +80,7 @@
   # Wrap the user's desired (unwrapped) Neovim package with arguments that'll be used to
   # generate a wrapped Neovim package.
   neovim-wrapped = inputs.mnw.lib.wrap {inherit pkgs;} {
-    appName = "nvf";
+    appName = config.vim.appname;
     neovim = config.vim.package;
     initLua = config.vim.builtLuaConfigRC;
     luaFiles = config.vim.extraLuaFiles;
--- a/modules/wrapper/rc/options.nix
+++ b/modules/wrapper/rc/options.nix
@@ -4,6 +4,19 @@
   inherit (lib.nvim.types) dagOf;
 in {
   options.vim = {
+    appname = mkOption {
+      type = str;
+      default = "nvf";
+      description = ''
+        Sets the {env}`NVIM_APPNAME` variable.
+
+        In traditional Neovim setups, standard directories can be further configured
+        by the {env}`NVIM_APPNAME` environment variable. This variable controls the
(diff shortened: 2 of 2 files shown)


horriblename:
>  this PR can be repurposed as a general-purpose wrapper refactoring one. Alternatively we should allow additional args to be passed to mnw.

the easiest solution is to just create a `vim.build.mnwOpts` and pass that to mnw to get the finalPackage, though I'm not sure what's worth exposing and what's not

in other news, I should swap vim.lazy.packages to use <https://gerg-l.github.io/mnw/options.html#plugins.optAttrs> and lift the stupid "attr name must equal package name" restriction

snoweuph (approved):
I like this.
just misses some changelog xD

Gerg-L:
why would you need a changelog entry for a new feature?

Gerg-L:
> Alternatively we should allow additional args to be passed to mnw. Maybe Gerg-L can chime in on that. Recursive merge? Just `//`?

@NotAShelf
do it as a submodule under `mnw.*`
consume this file: 
<https://github.com/Gerg-L/mnw/blob/master/modules/options.nix>
handle assertions: <https://github.com/Gerg-L/mnw/blob/0871dbf63a53610c95db04439ed8ea4d6ec9c160/modules/common.nix#L21-L25>
then pass the end result to `mnw.lib.uncheckedWrap`

should *just work*

the other option would be to convert nvf into a mnwModule and leave the current mnw options as-is and just wrap the mnw functions/modules with injecting nvf into them

snoweuph:
> why would you need a changelog entry for a new feature?

Because an Addition is also a change.
Like 80% of the `0.9` release is additions

NotAShelf:
> consume this file:
> [Gerg-L/mnw@`master`/modules/options.nix](https://github.com/Gerg-L/mnw/blob/master/modules/options.nix?rgh-link-date=2026-06-11T22%3A04%3A41Z)

seems a bit annoying to do tbh

Gerg-L:
@notashelf This could be closed, or made as a deferred option

NotAShelf:
I'll go with the latter.

PR #1635, opened 2026-06-11
lsp: fix deprecated diagnostic functions

### LINKED ISSUE #1585: deprecated functions vim.diagnostic.goto_prev and vim.diagnostic.goto_next
https://github.com/NotAShelf/nvf/blob/e86a92e4b29b499e5f1285b737b7612115103da9/modules/plugins/lsp/config.nix#L106

- ⚠️ WARNING vim.diagnostic.goto_prev() is deprecated. Feature will be removed in Nvim 0.13
  - ADVICE:
    - use vim.diagnostic.jump() instead.
    - stack traceback:
    
- ⚠️ WARNING vim.diagnostic.goto_next() is deprecated. Feature will be removed in Nvim 0.13
  - ADVICE:
    - use vim.diagnostic.jump() instead.
    - stack traceback:

horriblename:
fixes #1585


configuration.nix
```diff
@@ -68,7 +68,7 @@ isMaximal: {
       kotlin.enable = isMaximal;
       typescript.enable = isMaximal;
       go.enable = isMaximal;
-      lua.enable = isMaximal;
+      lua.enable = true;
```
NotAShelf: Move this up next to other default-enabled modules please.

modules/plugins/lsp/config.nix
```diff
@@ -103,8 +103,34 @@ in {
...
+              count = -1,
+              on_jump = function(_, bufnr)
+                vim.diagnostic.open_float({
+                  scope = "cursor",
+                  bufnr = bufnr,
+                  focus = false,
+                })
+              end,
+            })
+          end
```
NotAShelf, line 132: Should we have an internal Lua lib that exports  those with type hints etc. so that they may be reused?

NotAShelf (commented):
Two comments. Neither are blockers, but possibly worth addressing.

modules/plugins/lsp/config.nix
horriblename: uh I'm not really sure what you mean, this piece of code in particular isn't worth creating a new file for imo. If you're talking in general if we should export a lua lib, uhhh maybe? idk I'd need to think about it, and what should go into the lib

configuration.nix
horriblename: did not mean to do that, removed

snoweuph (approved):
LGTM

PR #1645, opened 2026-06-13
ci: add nvf cachix to linkcheck job

horriblename:
Added nvf.cachix.org to the final job taking forever to finish

Alternatively I could add nvf.cachix.org as substituter to every install-nix job. idk if there's any reason to


horriblename:
finally CI under 1min

NotAShelf (approved):

Issue #1180, opened 2025-10-12
No easy way to use LSP from PATH while falling back on a built-in one


miniluz:
### I have verified that this feature request has not been made before

- [x] I have checked the [issues tab](https://github.com/notashelf/nvf/issues?q=is%3Aissue), and did not find a similar feature request. I understand that my issue will be closed if it is a duplicate.

### Feature Type

Other

### Feature description

Right now, NVF hard-codes the versions of its LSPs from the pkgs that are sent in. As an example, https://github.com/NotAShelf/nvf/blob/main/modules/plugins/languages/rust.nix#L51 hardcodes the LSP command to be `pkgs.rust-analyzer`. I see the point of this - even if the system does not have rust-analyzer, nvim will come pre-packaged with one and will able to analyze Rust projects.

However, it does mean that if a project expects a different version of rust-analyzer, perhaps provided with a dev shell, nvim will _not_ use that version. My current project is an example of why this is an issue. A dev shell provides a newer version of rust-analyzer and the code gives weird errors if an older one is used. NVF by default gives me those errors because it's hard-coded to use an older version from nixpkgs.

I think the default behavior should be using the LSP on PATH and falling back to the built-in one, or alternatively it should be provided as an option so that it's easy to set up, because of the complexity of the current way I've found to do it (described on alternatives).

### Alternatives

To use the LSP on PATH, I can set the `languages.rust.lsp.package` option to `[ "rust-analyzer" ]`. But then neovim doesn't get a rust-analyzer built into it, and I lose the advantage of being able to use the LSP even on a system without Rust.

To fully get the behavior I want (PATH with a fallback) as a consumer, I could leave it as `[ "rust-analyzer" ]` and wrap my neovim appending `pkgs.rust-analyzer` to the PATH so that it uses it as a fallback. And that, as a solution, is not ergonomic and it'd probably be impossible to figure out users less experienced with Nix. Even more work would be needed to dynamically append the LSP to the PATH for various languages if they could be enabled or disabled, for example if I were using NVF as a NixOS or home-manager module.

sjcobb2022:
When 0.8 is released, this will be a lot easier since you can configure the command yourself. 

We do provide defaults, with the language modules, but those can be overriden like so:

vim.lsp.servers.rust.cmd = "rust-analyzer";

This will run the command simply as a bash script like you would set it up with regular lua.

theutz:
[...]

PR #1695, opened 2026-07-06
languages/haskell: respect existing binaries during HLS resolution

dathegreat:
Fixes a bug I encountered where NVF's pinned nixpkgs drifted far enough out of sync with one of my haskell project's pinned nixpkgs that the NVF-supplied HLS and my shell-supplied GHC stopped playing nicely together.

From the block comment I left in source:

>           HLS binaries are incredibly picky about which GHC version they run against
>           -- even failing between two GHC builds with identical version numbers, because one has a newer ABI.
>
>           Because of this, we shouldn't assume our provided HLS version is the right
>           one for a given project. Appending to PATH with `--suffix` takes advantage
>           of PATH's first-match lookup, so an existing HLS already in
>           the user's dev environment is found first and takes precedence over ours.
>           This way we provide a useful default without clobbering existing setups.

I can confirm that using `--suffix` no longer clobbered the HLS supplied by my dev shell, and all my haskell tooling worked as expected again.

## Sanity Checking

- Style and consistency
- Package(s) built:
  - [ ] `.#maximal`
- Tested on platform(s)
  - [x] `x86_64-linux`
  - [ ] `aarch64-linux`
  - [ ] `x86_64-darwin`
  - [ ] `aarch64-darwin`

---

### DIFF
--- a/docs/manual/release-notes/rl-0.9.md
+++ b/docs/manual/release-notes/rl-0.9.md
@@ -254,6 +254,9 @@
 - Added [csvview.nvim](https://github.com/hat0uma/csvview.nvim) support under
   `vim.utility.csvview` for rendering CSV/TSV files as aligned tables.
 
+- Switched from prepending -> appending our haskell-language-server binary to
+  PATH, so our binary is only used as a default / fallback.
+
 [alfarel](https://github.com/alfarelcynthesis):
 
 [obsidian.nvim]: https://github.com/obsidian-nvim/obsidian.nvim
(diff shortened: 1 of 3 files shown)


snoweuph:
btw you could also just modify `vim.languages.haskell.extensions.haskell-tools.setupOpts.hls.cmd = ["haskell-language-server" "--lsp"]` directly for your config, to not ship a hsl with your nvf config.

We generally prefer not using the `PATH` provided bins in NVF, but instead nix managed versions.
If people want to use the versions from their `PATH`, that is fine, but shouldn't really become default in nvf IMO.


@horriblename whats your opinion on this?

horriblename:
perhaps a good time to revisit https://github.com/NotAShelf/nvf/issues/1180#issue-3506794130 
(raf agreed to add a toggle for using either PATH or builtin, but I don't think we really talked about PATH+builtin fallback) 

suffixing goes against the rest of nvf so someone else might trip over that thinking the nvf hls has priority, so I don't think that's the way to go

I think i'd rather keep the current behavior till we add said toggle for #1180

PR #1792, opened 2026-08-19
docs/hacking: add guidelines for setupOpts options

horriblename:
Would like some feedback from other maintainers


docs/manual/hacking.md
```diff
@@ -732,6 +732,49 @@ own fields!
...
+
+For plugin modules that make use of the `setupOpts` pattern, try to keep the
+number of options you add to a minimum. You should only declare the options if
+one of these apply:
+
+1. The option benefits greatly from advanced type checking, and is commonly used
+   enough to justify the additional maintainer burden.
+2. It is used by some other nix module, usually for integration between
+   different plugins/modules.
+3. It is a path to some third-party program.
```
snoweuph, line 746: possibly add key mapping opts?

some plugins by default add keymaps without asking yk, that would basically doc them in nvf

snoweuph (dismissed):
I like it :3

docs/manual/hacking.md
```diff
@@ -607,7 +607,7 @@ Once the `npins` command is done, you can start referencing the plugin as a
 }
 ```
 
-### Packaging Complex Plugins {#sec-pkgs-for-plugins}
+### Packaging Plugins with a Special Build Step {#sec-pkgs-for-plugins}
```
NotAShelf:
```suggestion
### Packaging Plugins with a Special Build Steps {#sec-pkgs-for-plugins}
```

Singular kinda implies there cannot be more than one special step, which is not true for the Rust bullshit (I will dropkick whoever started the trend of loading Rust libs in Lua plugins)

docs/manual/hacking.md
```diff
@@ -732,6 +732,49 @@ own fields!
...
 
+#### Which Options to Add In `setupOpts` {#sec-setupopts-which-options-to-add}
+
+For plugin modules that make use of the `setupOpts` pattern, try to keep the
+number of options you add to a minimum. You should only declare the options if
+one of these apply:
+
+1. The option benefits greatly from advanced type checking, and is commonly used
+   enough to justify the additional maintainer burden.
+2. It is used by some other nix module, usually for integration between
```
NotAShelf:
```suggestion
2. It is used by some other Nix module, usually for integration between
```

docs/manual/hacking.md
```diff
@@ -732,6 +732,49 @@ own fields!
...
 ```
 
+#### Which Options to Add In `setupOpts` {#sec-setupopts-which-options-to-add}
+
+For plugin modules that make use of the `setupOpts` pattern, try to keep the
+number of options you add to a minimum. You should only declare the options if
+one of these apply:
+
+1. The option benefits greatly from advanced type checking, and is commonly used
+   enough to justify the additional maintainer burden.
```
NotAShelf: I think you should also mention that we add the option if it has any values worth documenting, e.g., the `example` field or just `documentation`.

docs/manual/hacking.md
```diff
@@ -732,6 +732,49 @@ own fields!
...
+{
+  options.vim.fzf-lua = {
+    enable = mkEnableOption "fzf-lua";
+    setupOpts = mkPluginSetupOption "fzf-lua" {
+      # 3rd case, program path
+      fzf_bin = mkOption {
+        type = str;
+        default = "${lib.getExe pkgs.fzf}";
+        description = "Path to fzf executable";
+      };
```
NotAShelf, line 766:
```suggestion
      fzf_bin = mkOption {
        type = str;
        default = "${lib.getExe pkgs.fzf}";
        example = "fzf"; # to get fzf from PATH
        description = "Path to fzf executable";
      };
```

Makes for a better example imo.

docs/manual/hacking.md
```diff
@@ -732,6 +732,49 @@ own fields!
...
+      # 1st and 2nd case,
+      # 1. `borderType` is a custom type we defined in our lib that is shared
+      #    by other plugins
+      # 2. Integrates with the vim.ui.borders module
+      winopts.border = mkOption {
+        type = borderType;
+        default = config.vim.ui.borders.globalStyle;
```
NotAShelf, line 773:
```suggestion
        default = config.vim.ui.borders.globalStyle;
        defaultText = lib.literalExpression "config.vim.ui.borders.globalStyle";
```

Causes extra eval during documentation build otherwise.

NotAShelf (changes requested):

docs/manual/hacking.md
horriblename: do you mean generalize the first point to "moar documentation" or add an entirely new point? I think it's close enough to generalize the first

